PR #90, opened 2015-01-28
feat(images): delete images from s3 when asked to

seanmonstar:
Lacking tests

### DIFF
--- a/lib/db/mysql.js
+++ b/lib/db/mysql.js
@@ -95,7 +95,8 @@ const Q_AVATAR_LIST = 'SELECT avatars.id, url, avatar_selected.avatarId '
 const Q_AVATAR_DELETE = 'DELETE FROM avatars WHERE id=?';
 
 const Q_PROVIDER_INSERT = 'INSERT INTO avatar_providers (name) VALUES (?)';
-const Q_PROVIDER_GET = 'SELECT * FROM avatar_providers WHERE name=?';
+const Q_PROVIDER_GET_BY_NAME = 'SELECT * FROM avatar_providers WHERE name=?';
+const Q_PROVIDER_GET_BY_ID = 'SELECT * FROM avatar_providers WHERE id=?';
 
 function firstRow(rows) {
   return rows[0];
@@ -129,7 +130,7 @@ MysqlStore.prototype = {
     id = buf(id);
     uid = buf(uid);
     var store = this;
-    return this.getProvider(provider).then(function(prov) {
+    return this.getProviderByName(provider).then(function(prov) {
       if (!prov) {
         throw AppError.unsupportedProvider(url);
       }
@@ -164,8 +165,12 @@ MysqlStore.prototype = {
     return this._write(Q_PROVIDER_INSERT, [name]);
   },
 
-  getProvider: function getProvider(name) {
-    return this._readOne(Q_PROVIDER_GET, [name]);
+  getProviderByName: function getProviderByName(name) {
+    return this._readOne(Q_PROVIDER_GET_BY_NAME, [name]);
+  },
+
+  getProviderById: function getProviderById(id) {
+    return this._readOne(Q_PROVIDER_GET_BY_ID, [id]);
   },
 
   _write: function _write(sql, params) {
--- a/lib/img/aws.js
+++ b/lib/img/aws.js
@@ -45,6 +45,29 @@ AwsDriver.prototype = {
         }
       });
     });
+  },
+
+  delete: function awsDelete(key) {
+    var s3 = this._s3;
+    var bucket = PUBLIC_BUCKET;
+    return new P(function(resolve, reject) {
+      logger.debug('delete.start', { bucket: bucket, key: key });
+      s3.deleteObjects({
+        Bucket: bucket,
+        Delete: {
+          Objects: [
+            { Key: key }
+          ]
+        }
+      }, function(err, data) {
+        if (err) {
+          reject(err);
+        } else {
+          logger.debug('delete.end', { key: key, data: data });
+          resolve(key);
+        }
+      });
+    });
   }
 
 };
--- a/lib/img/local.js
+++ b/lib/img/local.js
@@ -43,6 +43,25 @@ LocalDriver.prototype = {
         }
       });
     });
+  },
+
+  delete: function localDelete(key) {
+    return new P(function deletePromise(resolve, reject) {
+      var dir = PUBLIC_DIR;
+      logger.debug('delete.start', key);
+      if (/[^a-zA-Z0-9\_\-]/.test(key)) {
+        return reject(new Error('Illegal characters in key'));
+      }
+      fs.unlink(path.join(dir, key), function(err) {
+        if (err) {
+          logger.error('delete', err);
+          reject(err);
+        } else {
+          logger.debug('delete.end', key);
+          resolve(key);
+        }
+      });
+    });
   }
 
 };
--- a/lib/routes/avatar/delete.js
+++ b/lib/routes/avatar/delete.js
@@ -4,12 +4,40 @@
 
 const hex = require('buf').to.hex;
 const Joi = require('joi');
+const P = require('../../promise');
 
 const AppError = require('../../error');
+const config = require('../../config');
 const db = require('../../db');
+const logger = require('../../logging')('routes.avatar.delete');
+const request = require('../../request');
 const validate = require('../../validate');
 
+const WORKER_URL = config.get('worker.url');
 const EMPTY = Object.create(null);
+const FXA_PROVIDER = 'fxa';
+
+function workerDelete(id) {
+  return new P(function(resolve, reject) {
+    var url = WORKER_URL + '/a/' + id;
+    var opts = { method: 'delete', json: true };
+    logger.verbose('workerDelete', url);
+    request(url, opts, function(err, res, body) {
+      if (err) {
+        logger.error('network.error', err);
+        return reject(AppError.processingError(err));
+      }
+      if (res.statusCode >= 400 || body.error) {
+        logger.error('worker.error', body);
+        reject(AppError.processingError(body));
+        return;
+      }
+
+      logger.verbose('worker', body);
+      resolve();
+    });
+  });
+}
 
 function empty() {
   return EMPTY;
@@ -28,15 +56,25 @@ module.exports = {
         .required()
     }
   },
-  handler: function avatar(req, reply) {
+  handler: function deleteAvatar(req, reply) {
     db.getAvatar(req.params.id)
       .then(function(avatar) {
+        logger.debug('avatar', avatar);
         if (!avatar) {
           throw AppError.notFound();
         } else if (hex(avatar.userId) !== req.auth.credentials.user) {
           throw AppError.unauthorized('Avatar not owned by user');
         } else {
-          return db.deleteAvatar(req.params.id);
+          return P.all([
+            db.deleteAvatar(req.params.id),
+            db.getProviderById(avatar.providerId)
+          ]);
+        }
+      })
+      .spread(function(_, provider) {
+        logger.debug('provider', provider);
+        if (provider.name === FXA_PROVIDER) {
+          return workerDelete(req.params.id);
         }
       })
       .then(empty)
--- a/lib/server/worker.js
+++ b/lib/server/worker.js
@@ -8,6 +8,7 @@ const Joi = require('joi');
 const AppError = require('../error');
 const compute = require('../compute');
 const config = require('../config').root();
+const img = require('../img');
 const logger = require('../logging')('server.worker');
 
 exports.create = function() {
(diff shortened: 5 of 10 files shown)


lib/server/worker.js
```diff
@@ -52,6 +52,16 @@ exports.create = function() {
     }
   });
 
+  server.route({
+    method: 'DELETE',
+    path: '/a/{id}',
+    config: {
+      handler: function delete(req, reply) {
+        img.delete(req.params.id).done(reply, reply);
```
pdehaan: Not sure if I'm reading this right... but does this mean I can delete _**any**_ avatar from S3 if I know it's ID? Or is there logic in there to limit it to only being able to delete your own profile image for the currently logged in user?

seanmonstar: The worker is only accessible over the local network. So the only person who can tell the worker to delete something is the Profile web server.

lib/img/local.js
```diff
@@ -43,6 +43,22 @@ LocalDriver.prototype = {
         }
       });
     });
+  },
+
+  delete: function localDelete(key) {
+    return new P(function deletePromise(resolve, reject) {
+      var dir = PUBLIC_DIR;
+      logger.debug('delete.start', key);
+      fs.unlink(path.join(dir, name), function(err) {
```
ckarlof: What is `name` here? Is that supposed to be `key`? If so, are we relying on Hapi to do validation of `key` so it doesn't contain bad stuff like `../`?

ckarlof: Looks like some copy-paste error. Why does `upload` use `name` and `delete` use `key` as a parameter name?

lib/img/aws.js
```diff
@@ -45,6 +45,29 @@ AwsDriver.prototype = {
         }
       });
     });
+  },
+
+  delete: function awsDelete(key) {
+    var s3 = this._s3;
```
ckarlof: In terms of testing, can we mock the s3 object and verify that `deleteObjects` is called on it when `delete` is invoked?

lib/server/worker.js
ckarlof: @pdehaan, authentication was already handled [here](https://github.com/mozilla/fxa-profile-server/pull/90/files#diff-0ae6eb352cb79b1df84aa948c95de70dR36), but it's a good point that if that's every dropped for any reason, this would be problematic. 

Tests to the rescue.

ckarlof:
In this PR, actual images are only deleted when the `delete` endpoint is invoked. Since we're currently only allowing a single avatar per user, we should probably also delete the old avatar image if the user changers her avatar, since this is currently the only signal we have to reliers that "shit's changed".

ckarlof:
Overall, f+, but it needs tests with mocked s3 and filesystems objects, particularly to make sure don't regress the authentication issue @pdehaan raised.

lib/routes/avatar/delete.js
```diff
@@ -39,6 +39,9 @@ module.exports = {
           return db.deleteAvatar(req.params.id);
         }
       })
+      .then(function() {
+        return img.delete(req.params.id);
```
pdehaan: `grunt lint` says:
> 'img' is not defined.

lib/server/worker.js
pdehaan: `grunt lint` says:

> 'img' is not defined.

lib/img/local.js
```diff
@@ -43,6 +43,22 @@ LocalDriver.prototype = {
...
+        if (err) {
+          logger.error('delete', err);
+          reject(err);
+        } else {
+          logger.debug('delete.end', key);
+          resolve(key);
+        }
+      })
+    })
```
pdehaan: Missing semi-colons on lines 60-61, per `grunt lint`.

lib/server/worker.js
```diff
@@ -52,6 +52,16 @@ exports.create = function() {
     }
   });
 
+  server.route({
+    method: 'DELETE',
+    path: '/a/{id}',
+    config: {
+      handler: function delete(req, reply) {
```
pdehaan: `grunt lint` says:

> _Expected an identifier and instead saw 'delete' (a reserved word)._

Not sure if that matters (unless you're IE8 or something), but we may want to relax that JSHint warning. Plus, I/we should look into whether there the `grunt lint` task is, in fact, being run on Travis.

seanmonstar:
@ckarlof updated with mocks for S3. Other fixes likely fixed as well, siince tests were able to run completely.

Regarding deleting when a user changes avatars: The profile server was designed to allow a user to upload multiple avatars, and select 1 as their current selected. It wouldn't be possible to delete the image when they changed the selected one, unless they also delete that image.

It was designed this way to support the content-server. If those features aren't in use, one solution may be for the content-server to request deletion of the older image upon changing.

ckarlof:
> Regarding deleting when a user changes avatars: The profile server was designed to allow a user to upload multiple avatars, and select 1 as their current selected. It wouldn't be possible to delete the image when they changed the selected one, unless they also delete that image.

I'll talk with @zaach about this tomorrow.

ckarlof:
@rfk, can you take a look please?

rfk:
I should get a chance to look at this tomorrow morning.  @seanmonstar any idea what's up with the travis failure?

seanmonstar:
@rfk yep, there was some inconsistencies in the mysql API, I've fixed them up.

zaach:
>  If those features aren't in use, one solution may be for the content-server to request deletion of the older image upon changing.

We could do that for now.

lib/db/mysql.js
```diff
@@ -95,7 +95,8 @@ const Q_AVATAR_LIST = 'SELECT avatars.id, url, avatar_selected.avatarId '
 const Q_AVATAR_DELETE = 'DELETE FROM avatars WHERE id=?';
 
 const Q_PROVIDER_INSERT = 'INSERT INTO avatar_providers (name) VALUES (?)';
-const Q_PROVIDER_GET = 'SELECT * FROM avatar_providers WHERE name=?';
+const Q_PROVIDER_GET_NAME = 'SELECT * FROM avatar_providers WHERE name=?';
+const Q_PROVIDER_GET_ID = 'SELECT * FROM avatar_providers WHERE id=?';
```
rfk: nit: I would suggest `Q_PROVIDER_GET_BY_NAME`/`Q_PROVIDER_GET_BY_ID` here to match the naming of the functions that use them.  The current naming could ambiguously suggest that we're "getting the name" rather than "getting by name".

lib/img/local.js
```diff
@@ -43,6 +43,22 @@ LocalDriver.prototype = {
         }
       });
     });
+  },
+
+  delete: function localDelete(key) {
+    return new P(function deletePromise(resolve, reject) {
```
rfk: style nit: use of a named function for the closure here is inconsistent with other files in this repo; dunno whether you care or not but it was odd enough that I noticed :-)

lib/routes/avatar/delete.js
```diff
@@ -4,12 +4,40 @@
...
+const request = require('../../request');
 const validate = require('../../validate');
 
+const WORKER_URL = config.get('worker.url');
 const EMPTY = Object.create(null);
+const FXA_PROVIDER = 'fxa';
+
+function workerDelete(id) {
+  return new P(function(resolve, reject) {
+    var url = WORKER_URL + '/a/' + id;
```
rfk: There's now two places where we have "build a worker URL and make a request to it logic"; we should consider a follow-up refactor to better encapsulate some of this shared logic.

lib/img/local.js
seanmonstar: I believe I was hitting an error, and wanted to improve the stack trace. My preference is to name as many functions as I can, because it doesn't hurt and only ever improves stack traces...

rfk: :+1:

test/api.js
```diff
@@ -296,46 +308,100 @@ describe('/avatar', function() {
...
-        headers: {
-          authorization: 'Bearer ' + tok,
-        }
-      }).then(function(res) {
-        assert.equal(res.statusCode, 200);
-        return db.getAvatar(id);
-      }).then(function(avatar) {
-        assert.equal(avatar, undefined);
+
+      it('should remove avatar from user and s3', function() {
```
rfk: suggest "should remove avatar from db and s3"

test/api.js
```diff
@@ -296,46 +308,100 @@ describe('/avatar', function() {
...
+          url: '/avatar/' + id,
+          headers: {
+            authorization: 'Bearer ' + tok,
+          }
+        }).then(function(res) {
+          assert.equal(res.statusCode, 200);
+          return db.getAvatar(id);
+        }).then(function(avatar) {
+          assert.equal(avatar, undefined);
+          return Static.get(s3url);
```
rfk: IIUC, this is the part where we check whether the image has been deleted from the mocked S3.  We should add a check in the preceeding test to ensure that the mock worked correctly, i.e. check that we can `Static.get(s3url)` and receive a `200 OK`, to distinguish between "it was deleted successully" and "it was never uploaded in the first place".

test/lib/mock.js
```diff
@@ -121,5 +136,40 @@ module.exports = function mock(options) {
...
+        filter: function(record) {
+          if (cb(record)) {
+            log.removeFilter(filter);
+            return false;
+          }
+          return true;
+        }
+      };
+      log.addFilter(filter);
+    }
```
rfk: Nice, I like the use of logging output as an additional check in the tests.

test/lib/mock.js
```diff
@@ -76,7 +76,22 @@ module.exports = function mock(options) {
...
+      .post(u)
+      .reply(200, function(uri, body) {
+        var id = body.match(/<Key>([0-9a-f]{32})<\/Key>/)[1];
+        var s = through();
+        s.setEncoding = function() {};
+        local.delete(id).done(function() {
+          s.end();
+        });
+        return s;
+      });
```
rfk: It took me a good long while to figure out how all the pieces fit together here, and I'm not that confident I understood it all.  Does this basically mock out requests to S3 and replace them with corresponding calls to the local image backend?  It may be worth a comment just to orient new readers.

lib/img/aws.js
```diff
@@ -45,6 +45,29 @@ AwsDriver.prototype = {
         }
       });
     });
+  },
+
+  delete: function awsDelete(key) {
```
rfk: Echoing @pdehaan comments earlier, I don't think it'd hurt to sanity-check the key in the backends here, e.g. with a quick regex test to ensure it's alphanumeric.  In theory it's already validated at higher levels, but...

test/api.js
rfk: More generally, if we ask `nock` to mock out a route, can we somehow assert that that mocked route was actually triggered?

test/lib/mock.js
seanmonstar: Yes, exactly. The `nock.reply` method expects either a String or a Stream returned. Since the local driver is still async, I'm returning an empty stream that finishes once the local driver is complete.

lib/routes/avatar/delete.js
seanmonstar: indeed

test/api.js
seanmonstar: 1. It does: https://github.com/mozilla/fxa-profile-server/pull/90/files#diff-9265a69d5e81bfaab4edbd417a33a626L272
2. It appears there is a way, yes. The object returned by nock has `done` method that asserts all the mocks were triggered...

rfk: > 1. It does: 

Ah, nice.  And I guess this triggers the existing `uploadAws` mock for the S3 case as well.

> 2. It appears there is a way, yes. 

This sounds excellent :-)

lib/img/aws.js
seanmonstar: huh? The key could be anything, especially with multiple image sizes, there will be suffixes added.

The concern was here?

rfk: Accidentally passing through malicious keys with e.g. `..` or `/` or whatever in them; maybe less of a concern with the s3 backend though

seanmonstar: Ohhhh, ok. I've added a regex check into the Local driver.

seanmonstar:
@rfk addressed.

lib/img/aws.js
rfk: Commenting for completeness: I think this is sufficiently covered by the test that mocks s3 at the network level

test/api.js
```diff
@@ -34,10 +34,16 @@ const db = require('../lib/db');
...
 
+var imagePath = path.join(__dirname, 'lib', 'firefox.png');
+var imageData = fs.readFileSync(imagePath);
 
 const GRAVATAR =
   'http://www.gravatar.com/avatar/00000000000000000000000000000000';
 
+afterEach(function() {
+  mock.done();
+});
```
rfk: nice!

rfk:
r+ from me

PR #91, opened 2015-02-09
chore(install): check that `gm` is installed

### LINKED ISSUE #76: Be noisier if graphicsmagick isn't installed
The server started and installed fine without it, but the profile server will return 500s if the avatar API attempts to process any images.

seanmonstar:
Fixes #76

### DIFF
--- a/scripts/check_gm.sh
+++ b/scripts/check_gm.sh
@@ -1,0 +1,8 @@
+#!/bin/sh
+
+if hash gm 2>/dev/null; then
+  exit 0
+else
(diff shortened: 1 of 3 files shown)


scripts/check_gm.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+if hash gm 2>/dev/null; then
```
rfk: TIL about this `hash` command...

rfk:
:+1:

PR #92, opened 2015-02-09
test(mocks): improve state of mocking in tests

seanmonstar:


test/api.js
```diff
@@ -324,18 +250,15 @@ describe('/avatar', function() {
...
-      mockWorker();
-      mockAws();
+      });
+      mock.image();
       return Server.api.post({
         url: '/avatar/upload',
-        payload: imageData,
-        headers: {
-          authorization: 'Bearer ' + tok,
+        payload: imageData, headers: { authorization: 'Bearer ' + tok,
```
rfk: why remove newline here?

rfk:
Otherwise, this looks great and certainly cleans up the test file nicely.

test/api.js
seanmonstar: I... didn't even notice this...

PR #95, opened 2015-02-18
feat(avatar): add multiple image support

### LINKED ISSUE #68: Add ability to return device appropriate profile image sizes
Currently it looks like we only generate and return a 600-by-600px (264kb, in my case) profile image (see https://firefoxusercontent.com/e78698b9ace433dedd6c93c9574b0d4d)

If I go to https://accounts.firefox.com/settings the avatar size is 120-by-120px, and on https://accounts.firefox.com/settings/avatar it is 240-by-240px.

It would be lovely if we had some API to arbitrarily request a random image size (a la [gravatar](https://en.gravatar.com/site/implement/images/#size)), or maybe we pre-generate common sizes of avatars that we support (ie: 100x100, 120x120, 200x200, 240x240, 300x300, 600x600, etc).

That way if I only want to display a 200px profile image, I can only download 60kb instead of a 260kb image that gets squashed client side).

#mo-px-mo-problems

### LINKED ISSUE #89: Profile Image Sizes
Meta bug for shipping multiple image sizes to production.

- [ ] [Add multiple sizes to profile server](https://github.com/mozilla/fxa-profile-server/issues/68)
- [x] Document additional image sizes
- [x] Improve testing framework to properly unit test

seanmonstar:
When an avatar is uploaded, we will now generate multiple sizes of the
image, to better accommodate various environments that the avatar may
appear.

The sizes generated are based of current constants in the `lib/img/index.js`
module. The `SIZES` constant contains a map of `suffix` to `height/width`
values. The suffix `default` is special cased, and will actually mean no
suffix. Every other value will be appended on the end of the URL, with an
underscore.

The current `SIZES` are:

    {
      small: { h: 100, w: 100 },
      default: { h: 200, w: 200 },
      large: { h 600, w: 600 }
    }

which generate 3 images when an upload occurs:

- https://s3.url/<imageid>_small (100px x 100px)
- https://s3.url/<imageid> (200px x 200px)
- https://s3.url/<imageid>_large (600px x 600px)

Closes #68
Closes #89

### DIFF
--- a/lib/compute/index.js
+++ b/lib/compute/index.js
@@ -13,6 +13,7 @@ const logger = require('../logging')('compute');
 const P  = require('../promise');
 
 const MAX_PROCESSES = config.get('img.compute.maxProcesses');
+const SIZES = require('../img').SIZES;
 
 var imageCc;
 if (MAX_PROCESSES > 0) {
@@ -43,6 +44,19 @@ if (MAX_PROCESSES > 0) {
   };
 }
 
+function enqueue(msg) {
+  return new P(function enqueuePromise(resolve, reject) {
+    imageCc.enqueue(msg, function(err, res) {
+      if (err) {
+        logger.error('process.error', err);
+        reject(AppError.processingError(err));
+      } else {
+        resolve(res);
+      }
+    });
+  });
+}
+
 exports.image = function image(id, payload) {
   return new P(function(resolve, reject) {
     toArray(payload, function(err, arr) {
@@ -50,17 +64,17 @@ exports.image = function image(id, payload) {
         return reject(err);
       }
       var buf = Buffer.concat(arr);
-      imageCc.enqueue({
-        id: id,
-        payload: buf
-      }, function(err, res) {
-        if (err) {
-          logger.error('process.error', err);
-          reject(AppError.processingError(err));
-        } else {
-          resolve(res);
-        }
-      });
+
+      resolve(P.all(Object.keys(SIZES).map(function(variant) {
+        var size = SIZES[variant];
+        return enqueue({
+          id: id,
+          suffix: (variant === 'default') ? '' : variant,
+          height: size.h,
+          width: size.w,
+          payload: buf
+        });
+      })));
     });
   });
 };
--- a/lib/server/worker.js
+++ b/lib/server/worker.js
@@ -4,12 +4,15 @@
 
 const Hapi = require('hapi');
 const Joi = require('joi');
+const P = require('../promise');
 
 const AppError = require('../error');
 const compute = require('../compute');
 const config = require('../config').root();
 const img = require('../img');
 const logger = require('../logging')('server.worker');
+
+const SIZES = img.SIZES;
 
 exports.create = function() {
   var server = Hapi.createServer(
@@ -58,7 +61,12 @@ exports.create = function() {
     path: '/a/{id}',
     config: {
       handler: function delete_(req, reply) {
-        img.delete(req.params.id).done(reply, reply);
+        P.all(Object.keys(SIZES).map(function(name) {
+          if (name === 'default') {
+            return req.params.id;
+          }
+          return req.params.id + '_' + name;
+        })).map(img.delete).done(reply, reply);
       }
     }
   });
--- a/docs/API.md
+++ b/docs/API.md
@@ -136,6 +136,12 @@ Returns details of the current user avatar, or an empty object if none.
 
 An avatar `id` is a 32-length hexstring.
 
+All avatars hosted by Firefox Accounts (see 3rd-party provider docs for
+their equivalent) can be accessed as multiple sizes. The default size is
+200x200 pixels. There is are small (100x100) and large (600x600)
+variants, which can accessed by adding the `_small` or `_large` suffix
+to the avatar URL.
+
 #### Request
 
 ```sh
--- a/lib/compute/image-cc.js
+++ b/lib/compute/image-cc.js
@@ -9,15 +9,12 @@ const img = require('../img');
 const logger = require('../logging')('compute.image-cc');
 const P = require('../promise');
 
-const HEIGHT = String(config.get('img.resize.height'));
-const WIDTH = String(config.get('img.resize.width'));
 const CONTENT_TYPE_PNG = 'image/png';
 
 logger.info('worker.config', { config: config.get('img') });
 
-function processImage(src) {
+function processImage(src, width, height) {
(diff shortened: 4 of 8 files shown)


rfk:
> The sizes generated are configurable.

This is cool, but what happens if/when we tweak this configuration?  How do consumers of these images know what sizes are available, and whether this will vary per user?

seanmonstar:
@rfk It probably shouldn't be changed. There's no mechanism to "upgrade" or "downgrade" to a different configuration. It currently exists just so UX can decide what sizes to have, independent of this PR being mergeable.

rfk:
Will we batch-process the (I assume small number of) avatars that are currently in the system so that they're available in these sizes?

rfk:
Do we need to update e.g. the API doc with instructions on accessing the different sizes, or is this covered elsewhere?

lib/compute/index.js
```diff
@@ -50,17 +64,17 @@ exports.image = function image(id, payload) {
...
+      resolve(P.all(Object.keys(SIZES).map(function(variant) {
+        var hw = SIZES[variant];
+        return enqueue({
+          id: id,
+          suffix: (variant === 'default') ? '' : variant,
+          height: hw.h,
+          width: hw.w,
+          payload: buf
+        });
+      })));
```
rfk: Do we need any special error-handling logic here.  For example, if some sizes get processes correctly and some error out due to e.g. compute-cluster queue filling up, should we try to clean up after ourselves?

rfk:
The code here looks good apart from question around error handling, but I'd like to see the broader docs/plan for how this will be exposed to consumers of the API.

lib/compute/index.js
seanmonstar: Probably (gross).

seanmonstar:
1. Currently lacking some way to bulk update this. We could have a tool that tries to create image sizes for all images lacking any sizes (and thus usable if we ever add more sizes), or just bulk create the ones currently missing with a shotgun.

2. The docs should be updated with the sizes/suffixes available. This can be done as soon as we know what sizes we actually should have.

  That this is a config value doesn't mean the config should be adjusted on a whim. The values that we use should likely **be** the default config values, and the docs kept in sync. It's config mostly because the code is generic enough.

lib/compute/index.js
rfk: Gross indeed, but I think worthwhile.  It may require extract some higher-level "task" abstraction here to keep things sane.

rfk:
> That this is a config value doesn't mean the config should be adjusted on a whim.
> The values that we use should likely be the default config values, and the docs kept in sync.
> It's config mostly because the code is generic enough.

I dunno, something about this makes me nervous.  It opens the possibility for people to tweak without understanding the consequences, and it adds noise when trying to understand what our actual configuration knobs are.  What do you think about just making it a module-level constant rather than exposing it to config?

rfk:
A broader question is, how does this relate to other avatar providers such as gravatar?  Short of inspecting the image URL, it doesn't seem like the relier can determine what provider is hosting the image, and hence how they should ask for different sizes.

It would be good if reliers didn't have to worry about the details of different providers, but I'm not sure the best way to achieve that in our API.

seanmonstar:
> What do you think about just making it a module-level constant rather than exposing it to config?

Ok, I'll pull it out of the config, and keep them as constants.

> A broader question is, how does this relate to other avatar providers such as gravatar?

Gravatar allows passing a `?s=200` query string to set the size value. I had originally wanted something similar, but @ckolos worried about being able to map those URLs onto S3/Cloudfront. Ideally, I'd map `?s=100` or something to `_small`, and so forth, so we can offer something people are more used to.

lib/compute/index.js
seanmonstar: I dug into the S3 docs, and there is no sort of "transaction" for uploading multiple files, only multiple parts of a single file. It seems to properly accomplish this, we would need to collect any and all errors uploading, and try to delete all images that didn't error. Those delete requests would also need to be monitored, and perhaps retried (up to a maximum?).

*shudder*

seanmonstar:
@rfk I've updated to no longer use the config to manage this; it's now a constant in the `img` module. Updated docs and commit message as well.

lib/compute/index.js
rfk: Yeah, the more I think about all the edge cases here, the less I want to tackle it.  We can always spin off cleanup to a follow-up bug.

lib/compute/index.js
```diff
@@ -50,17 +64,17 @@ exports.image = function image(id, payload) {
...
-        if (err) {
-          logger.error('process.error', err);
-          reject(AppError.processingError(err));
-        } else {
-          resolve(res);
-        }
-      });
+
+      resolve(P.all(Object.keys(SIZES).map(function(variant) {
+        var hw = SIZES[variant];
```
rfk: nit: `size` or even `sz` might be a clearer name here than `hw`, which to me reads either "hardware" or "high-water" on first glance.

lib/server/worker.js
```diff
@@ -58,7 +61,12 @@ exports.create = function() {
     path: '/a/{id}',
     config: {
       handler: function delete_(req, reply) {
-        img.delete(req.params.id).done(reply, reply);
+        P.all(Object.keys(SIZES).map(function(name) {
+          if (name === 'default') {
+            return req.params.id;
+          }
+          return req.params.id + '_' + name;
+        })).map(img.delete).done(reply, reply);
```
rfk: Given the possibility for failures in both upload and in cleanup, I think we should translate a `404 Not Found` failure on delete into "sure, yeah, I deleted that, no errors here" in this API.

rfk: Actually "we should" is too strong; but we should consider it.

rfk:
Nit: the commit message "add multiple image support" doesn't seem to capture what's going on here as it suggests different images.  Something like "add support for multiple image sizes" may be clearer.

rfk:
I'm not opposed to using different API approach between gravatar and our own hosting, but we should make it easy for the client to distinguish which provider is in use.  Can we do better than "check the domain name in the URL" here?  We do it this way internally but the regexes involved are...not exactly pleasant :-)

rfk:
@seanmonstar do you want to land the processing code in order to get it out and running in prod, and iterate on the client-facing API as a follow-up?

lib/compute/index.js
seanmonstar: Changed to `size`

PR #99, opened 2015-03-11
fix(docs): note "avatar" field in /v1/profile response

rfk:
I noticed while working on https://github.com/mozilla/PyFxA/pull/15 that the `/v1/profile` endpoint returns an "avatar" field, but this isn't shown in the example in docs.


rfk:
So yeah...I'm pretty sure those failing travis builds are not the result of this one-line docs fix :-)

Issue #84, opened 2015-01-06
Support "display name" profile field


zaach:

zaach:
fixed in #103

PR #103, opened 2015-03-26
feat(display_name): add a display name field

zaach:
Note: this modifies the SQL schema, adding a new `profile` table with a displayName field.

Fixes issue #84

@rfk or @seanmonstar r?


lib/db/schema.sql
```diff
@@ -18,3 +18,8 @@ CREATE TABLE IF NOT EXISTS avatar_selected (
   avatarId BINARY(16) NOT NULL,
   FOREIGN KEY (avatarId) REFERENCES avatars(id) ON DELETE CASCADE
 ) ENGINE=InnoDB CHARACTER SET utf8 COLLATE utf8_unicode_ci;
+
+CREATE TABLE IF NOT EXISTS profile (
+  userId BINARY(16) NOT NULL PRIMARY KEY,
+  displayName VARCHAR(256)
```
zaach: What size should this be?

chilts:
Hey @zaach, here's a quick summary pointing to a few things we do for database patches in the auth and oauth servers.

In each repo, there are a set of (consecutive) patches here:

* https://github.com/mozilla/fxa-auth-db-mysql/tree/master/db/schema

We tell the server that it should expect this patch level to be set (this is set at the end of each patch). If the db isn't this patch level or above, it'll b0rk:

* https://github.com/mozilla/fxa-auth-db-mysql/blob/master/db/patch.js
* https://github.com/mozilla/fxa-auth-db-mysql/blob/master/db/mysql.js#L123-L127

In dev, we run this (local) db_patcher.js script which uses the `mysql-patcher` package:

* https://github.com/mozilla/fxa-auth-db-mysql/blob/master/bin/db_patcher.js
* https://github.com/mozilla/mysql-patcher/

In prod, we usually take the patch file and do it manually (at the moment). Some wheels are in place to also make this automatic, but we're not there yet. Here's an example:

* https://bugzilla.mozilla.org/show_bug.cgi?id=1091194

All this doesn't mean you have to do all that in this patch but it might help if you want to merge around the same technique we use in the auth and oauth servers. Any Q's let me know. :)

rfk:
The oauth server also uses this scheme, for another point of comparison:

  https://github.com/mozilla/fxa-oauth-server/tree/master/lib/db/mysql

(It automagically applies schema updates without use of a separate db_patcher.js script when in dev mode, which may be a better fit for workflow on this repo)

chilts:
Ah yeah, this is also automatic when `npm start` is run in the auth-db-mysql server but it's in a separate script rather than in the main server code.

test/api.js
```diff
@@ -478,6 +498,75 @@ describe('/avatars', function() {
...
+          headers: {
+            authorization: 'Bearer ' + tok
+          }
+        }).then(function(res) {
+          assert.equal(res.statusCode, 200);
+          assert.equal(JSON.parse(res.payload).displayName, 'Spock');
+        });
+      });
+    });
+
```
seanmonstar: Missing test case is when a display name has not been set.

What should happen? The code currently looks like it will return `{}` if it doesn't exist. Should it instead be a 404? Or stay empty?

zaach: I used `{}` to be consistent with `GET /avatar` when there's no avatar, but 404 would be more RESTy. We could break consistency here and use 404. Eventually we could update avatar and clients to handle 404 as well, or just leave it inconsistent. Or stay with an empty response and 200 status and be less RESTy. I'm not really opinionated on it, but maybe using a 404 is fine and we can live with the inconsistency.

zaach:
nalexander reminded me on IRC that we need some kind of way to signal to clients that the data hasn't been modified since they last fetched it, e.g. `If-unmodified-since` or etags.

zaach:
@chilts So, to bootstrap this for the profile server database I assume I should create the patch file for the `dbMetadata` table first (000), followed by the current schema (001), then finally the new `profile` table patch (002). Then for production, we can manually apply the metadata table patch and the new profile table patch? Will dev Just Work if I follow the setup in oauth-server?

chilts:
@zaach sorry, I missed this comment. Do you need anything else here apart from the review of the PR you've already done? :) /me scrambles back in time

zaach:
@chilts Dev seems to work fine, so that's resolved. I'm not clear on how this will affect production, although that's probably more of a nice-no-know for me– I assume ckolos et al know what to do.

chilts:
@zaach yes, for prod, we usually create separate patches which go into Bugzilla to be applied by ckolos separately, since we don't (currently) use mysql-patcher in prod. In this case, then you can just create the `dbMetadata` table, do your _new_ SQL statements and insert patch level 2 (if you like) and ignore those older tables that already exist.

seanmonstar:
Does this need a rebase since it included a commit from the other PR?

zaach:
yep, on it

PR #104, opened 2015-04-06
feat(mysql): use mysql patcher to allow incremental schema updates

zaach:
@seanmonstar or @chilts r?

### DIFF
--- a/lib/db/mysql/index.js
+++ b/lib/db/mysql/index.js
@@ -2,84 +2,102 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+const path = require('path');
 const mysql = require('mysql');
 const buf = require('buf').hex;
+const MysqlPatcher = require('mysql-patcher');
 
-const AppError = require('../error');
-const config = require('../config');
-const logger = require('../logging')('db.mysql');
-const P = require('../promise');
+const AppError = require('../../error');
+const config = require('../../config');
+const logger = require('../../logging')('db.mysql');
+const P = require('../../promise');
+const patch = require('./patch');
 
-const SCHEMA = require('fs').readFileSync(__dirname + '/schema.sql').toString();
 
 function MysqlStore(options) {
   if (options.charset && options.charset !== 'UTF8_UNICODE_CI') {
-    logger.warn('createDatabase', { charset:  options.charset });
+    logger.warn('createDatabase', { charset: options.charset });
   } else {
     options.charset = 'UTF8_UNICODE_CI';
   }
+  options.typeCast = function(field, next) {
+    if (field.type === 'TINY' && field.length === 1) {
+      return field.string() === '1';
+    }
+    return next();
+  };
   this._pool = mysql.createPool(options);
 }
 
-function createSchema(client, options) {
-  logger.verbose('createSchema', options);
+// Apply patches up to the current patch level.
+// This will also create the DB if it is missing.
+
+function updateDbSchema(patcher) {
+  logger.verbose('updateDbSchema', patcher.options);
 
   var d = P.defer();
-  var database = options.database;
+  patcher.patch(function(err) {
+    if (err) {
+      logger.error('updateDbSchema', err);
+      return d.reject(err);
+    }
+    d.resolve();
+  });
 
-  logger.verbose('createDatabase', database);
-  client.query('CREATE DATABASE IF NOT EXISTS ' + database
-    + ' CHARACTER SET utf8 COLLATE utf8_unicode_ci', function(err) {
-      if (err) {
-        logger.error('createDatabase', err);
-        return d.reject(err);
+  return d.promise;
+}
+
+// Sanity-check that we're working with a compatible patch level.
+
+function checkDbPatchLevel(patcher) {
+  logger.verbose('checkDbPatchLevel', patcher.options);
+
+  var d = P.defer();
+
+  patcher.readDbPatchLevel(function(err) {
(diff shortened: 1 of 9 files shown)


zaach:
Ergh. Looks like nodemon needs to be updated.

lib/db/mysql/index.js
```diff
@@ -2,85 +2,103 @@
...
+  patcher.readDbPatchLevel(function(err) {
+    if (err) {
+      logger.error('checkDbPatchLevel', err);
+      return d.reject(err);
+    }
+    // We are only guaranteed to run correctly if we're at the current
+    // patch level for this version of the code (the normal state of
+    // affairs) or the one immediately above it (during a deployment).
+    if (patcher.currentPatchLevel !== patch.level) {
+      if (patcher.currentPatchLevel !== patch.level + 1) {
```
chilts: In fxa-auth-db-mysql, we decided to go with it being okay if the DB is at any patch level greater than the one expected. We started with the same as you have here. Just mentioning it here, not expecting you to change it.

I think at one stage it made it easier if there were multiple patches in a release.

lib/db/mysql/index.js
```diff
@@ -2,85 +2,103 @@
...
+    logger.warn('createDatabase', { charset: options.charset });
   } else {
     options.charset = 'UTF8_UNICODE_CI';
   }
+  options.typeCast = function(field, next) {
+    if (field.type === 'TINY' && field.length === 1) {
+      return field.string() === '1';
+    }
+    return next();
+  };
```
chilts: I don't know what this `typeCast` thing is or what it's doing. It doesn't seem to be used (in this diff), is it meant to be here?

zaach: That was copy-pasted from the oauth-server's `mysql/index.js`. We can probably remove it here though.

lib/db/mysql/patches/patch-002-001.sql
```diff
@@ -0,0 +1,9 @@
+--  Drop all the tables.
+--  (commented out to avoid accidentally running this in production...)
+
+-- DROP TABLE avatar_providers;
+-- DROP TABLE avatars;
+-- DROP TABLE avatar_selected;
+
```
chilts: I usually try and make sure all statements in the reverse patch are the exact same reverse order as in the forward patch. This only makes sense when latter statements depend on earlier statements.

For example, does MySql complain here when `avatars` is dropped, but is still referenced from `avatar_selected`? And if not, it should! :)

chilts: (Except the `UPDATE dbMetadata` of course, which is always last.)

lib/db/mysql/index.js
zaach: It seems like this patcher code could be extracted into it's own module, since there's some duplication across repos now. Maybe for another time.

lib/db/mysql/index.js
```diff
@@ -2,85 +2,103 @@
...
+      return updateDbSchema(patcher);
+    }
+  }).then(function() {
+    return checkDbPatchLevel(patcher);
+  }).then(function() {
+    return P.promisify(patcher.end, patcher)();
+  }).then(function() {
+    return new MysqlStore(options);
+  });
 };
```
chilts: Not sure if you want a `.catch()` function here, just in case something goes wrong. Ah, perhaps it is checked from the calling function. Maybe not. I'm probably overlooking something.

```js
  if (config.get('db.driver') === 'mysql') {
    p = klass.connect(config.get('mysql'));
  } else {
    p = klass.connect();
  }
  return p.then(function(store) {
    logger.debug('connected', config.get('db.driver'));
    driver = store;
  }).then(loadProviders).then(function() {
    return driver;
  });
```

lib/db/mysql/patches/patch-001-002.sql
```diff
@@ -0,0 +1,30 @@
+--  Create the initial set of tables.
+--
+--  Since this is the first migration, we use `IF NOT EXISTS` to allow us
+--  to run this on a db that already has the original schema in place. The
+--  patch will then be a no-op.  Subsequent patches should *not* use `IF
+--  NOT EXISTS` but should fail noisily if the db is in an unexpected state.
+
+CREATE TABLE IF NOT EXISTS avatar_providers (
+  id INT AUTO_INCREMENT PRIMARY KEY,
```
chilts: Since we're grafting the mysql-patcher into the workflow here, the `IF NOT EXISTS` is fine here (ditto for the next two tables too). For reference, for future patches we wouldn't use it since we want each patch to be deterministic. Either it all worked, or something was wrong. So yeah, this is okay for this patch.

lib/db/mysql/index.js
chilts: Yeah, we're doing a similar thing in a few places. Maybe mysql-patcher could also do the check?

lib/db/mysql/index.js
seanmonstar: It's passed an option to `mysql.createPool`. By default, `mysql` treats `BOOLEAN` columns as `TINY`, and returns `1` or `0`. That works in most cases, when just testing truthiness, but it can cause problems when testing identity. So this function simply evaluates the field into a boolean instead.

chilts:
@zaach anything else you need from me for this one or are you all good with whatever you need to do with nodemon?

zaach:
@chilts Do we actually need nodemon? I'm inclined to remove it if no one is using it.

seanmonstar:
I used it for runnning `npm start` locally. However, I never do that anymore. I only ever run `npm test`. If no one else finds it useful, it can die.

PR #109, opened 2015-04-30
feat(profile): return all /profile pieces that scopes allow

### LINKED ISSUE #108: multiple fields with one request
We currently limit which fields third-party reliers can access, so they cannot use the open ended `profile` scope to get all fields in one request. They have to send separate requests for each field they want access to. It would be nice we supported returning multiple fields from a single request for scopes that the relier does have permission to use.

seanmonstar:
So basically, using Hapi's `server.inject()`, the `/profile` endpoint just clumps together the results of requests to each lesser piece of information. Any 403s (scope not allowed) and 404s (such as has no avatar) simply means that value isn't in the returned result.

Closes #108

### DIFF
--- a/lib/batch.js
+++ b/lib/batch.js
@@ -1,0 +1,58 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const logger = require('./logging')('batch');
+const P = require('./promise');
+
+function inject(server, options) {
+  return new P(function(resolve) {
+    server.inject(options, resolve);
+  });
+}
+
+// called in a route as `batch(req, map)`, where map is an Object
+// mapping property names to routes.
+//
+// Ex: batch(req, { email: '/v1/email', avatar: '/v1/avatar' }).done(reply)
+//
+// The above will fetch both routes as GET requests, into `emailRes` and
+// `avatarRes`. The end result of the returned promise would then be:
+//
+// { email: emailRes['email'], avatar: avatarRes['avatar'] }
+function batch(request, map) {
+  var result = {};
+  Object.keys(map).forEach(function(prop) {
+    var url = map[prop];
+    result[prop] = inject(request.server, {
+      method: 'get',
+      url: url,
+      headers: request.headers,
+      credentials: request.auth.credentials
+    }).then(function(res) {
+      switch (res.statusCode) {
+        case 200:
+          return res.result[prop];
+        case 403:
+        case 404:
+          logger.debug('batch.' + res.statusCode, {
+            scope: request.auth.credentials.scope,
+            response: res.result
+          });
+          return undefined;
+        default:
+          throw res.result;
+      }
+    });
+  });
+  return P.props(result).then(function(result) {
+    Object.keys(result).forEach(function(key) {
+      if (result[key] === undefined) {
+        delete result[key];
+      }
+    });
+    return result;
+  });
+}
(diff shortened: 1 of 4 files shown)


seanmonstar:
cc @rfk @zaach

lib/routes/profile.js
```diff
@@ -2,35 +2,35 @@
...
-          email: creds.email,
-          uid: creds.user,
-          avatar: avatar ? avatar.url : null,
-          displayName: profile && profile.displayName ?
-            profile.displayName : null
-        };
-      });
+    var scopes = (req.auth.credentials.scope || []).filter(function(scope) {
+      return scope === 'profile' || scope.indexOf('profile:') !== -1;
+    });
```
rfk: nit: this could terminate iteration after finding a profile-related scope, rather than returning all of them.  Not worth it if it would ugly up the code though.

rfk:
@seanmonstar nice approach!  AFAICT this won't even result in extra db traffic, since we had separate `db.getSelectedAvatar` and `db.getDisplayName` calls already.

lib/batch.js
```diff
@@ -0,0 +1,49 @@
...
+
+function batch(request, map) {
+  var result = {};
+  Object.keys(map).forEach(function(prop) {
+    var url = map[prop];
+    result[prop] = inject(request.server, {
+      method: 'get',
+      url: url,
+      headers: request.headers,
+      credentials: request.auth.credentials
```
rfk: Does passing this through here mean that we don't have to re-check the credentials on each sub-request?

lib/batch.js
```diff
@@ -0,0 +1,49 @@
...
+const logger = require('./logging')('batch');
+const P = require('./promise');
+
+function inject(server, options) {
+  return new P(function(resolve) {
+    server.inject(options, resolve);
+  });
+}
+
+function batch(request, map) {
```
rfk: A high-level comment on what this function does would be helpful and the expected structure of its arguments would be helpful.

rfk:
I haven't dug into what's up with travis, but once that's green, I'm r+ modulo nits on this.  Thanks for jumping on this so quickly @seanmonstar.

lib/batch.js
seanmonstar: Yea! I at first wasn't passing it, and was noticing an explosion of requests to the oauth server to verify the same token over and over. It turns out, passing this short circuits it. (This is how hapi's bassmaster implements a batch route, I learned from it.)

lib/routes/profile.js
seanmonstar: Sure. In fact, I added this as the last thing, when I noticed it never returned 403 otherwise. It felt like a hack to a rest of the system...

seanmonstar:
@rfk force-pushed some fixes

lib/routes/profile.js
```diff
@@ -2,35 +2,43 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+const Boom = require('hapi').error;
 const Joi = require('joi');
 
-const db = require('../db');
+const batch = require('../batch');
+
+function hasProfileScope(scopes) {
```
rfk: :heart:

rfk:
Travis bustage looks like we may be triggering a bug in the db patch handling?

rfk:
I wonder if it's a race-condition where multiple db.* calls are trying to patch up to latest version at the same time.

seanmonstar:
Interesting. I tried a thing, let's see how travis likes it.

rfk:
green!

PR #115, opened 2015-06-02
feat(profile): add etag to profile API endpoint

### LINKED ISSUE #112: Add etag to profile endpoint
This will allow clients to know if *any* profile data has changed since they last fetched it.

eoger:
Fixes #112

### DIFF
--- a/lib/routes/profile.js
+++ b/lib/routes/profile.js
@@ -4,6 +4,7 @@
 
 const Boom = require('hapi').error;
 const Joi = require('joi');
+const checksum = require('checksum');
 
 const batch = require('../batch');
 
@@ -14,6 +15,13 @@ function hasProfileScope(scopes) {
     if (scope === 'profile' || scope.indexOf('profile:') === 0) {
       return true;
     }
+  }
+  return false;
+}
+
+function computeEtag(profile) {
+  if (profile) {
+    return checksum(JSON.stringify(profile));
   }
   return false;
 }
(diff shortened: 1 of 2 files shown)


lib/routes/profile.js
```diff
@@ -18,6 +19,20 @@ function hasProfileScope(scopes) {
...
 
+function computeEtag(profile) {
+  if (profile) {
+    var key = profile.email;
+    if (profile.avatar && profile.avatar.id) {
+      key += profile.avatar.id;
+    }
+    if (profile.displayName) {
+      key += profile.displayName;
+    }
```
seanmonstar: To prevent additional fields from being forgotten, how about just checksuming the JSON? basically `return checksum(JSON.stringify(profile))`

eoger: Good idea, I'm on it.

test/api.js
```diff
@@ -148,6 +149,28 @@ describe('/profile', function() {
...
+          authorization: 'Bearer ' + tok
+        }
+      }).then(function(res) {
+        assert.equal(res.statusCode, 200);
+        var etag = res.headers.etag.substr(1, 40);
+        var expectedEtag = checksum('user@example.domain' + 'Spock');
+        assert.equal(etag, expectedEtag);
+      });
+    });
+  });
```
zaach: Let's also test that a change in profile data results in a new etag.

test/api.js
```diff
@@ -148,6 +149,28 @@ describe('/profile', function() {
...
+          authorization: 'Bearer ' + tok
+        }
+      }).then(function(res) {
+        assert.equal(res.statusCode, 200);
+        var etag = res.headers.etag.substr(1, 40);
+        var expectedEtag = checksum(JSON.stringify(res.result));
+        assert.equal(etag, expectedEtag);
+      });
+    });
+  });
```
zaach: Oops, lost my comment on an outdated diff:

> Let's also test that a change in profile data results in a new etag.

jrgm: I don't see where @zaach's comment above was addressed.

PR #116, opened 2015-06-10
chore(config): Update convict and switch on strict validation.

### LINKED ISSUE #113: set convict to strict validation
Stealing @seanmonstar's excellent https://github.com/mozilla/fxa-oauth-server/issues/263 issue

---

> _"This will catch when any config value is added that shouldn't have been, or typos. Requires a bump to convict@0.8"_

rfk:
Fixes #113

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -88,6 +88,10 @@ const conf = convict({
     level: {
       env: 'LOG_LEVEL',
       default: 'info'
+    },
+    debug: {
+      env: 'LOG_DEBUG',
+      default: false
     }
   },
   mysql: {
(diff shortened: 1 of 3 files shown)


rfk:
I think the travis fail here is npm bustage, it's telling me that it can't find some dependency that's clearly in npm-shrinkwrap.  I'll see if I can convince it to go green...

anyways, @chilts r?

rfk:
yeah, after re-pushing, a different subset of tests fail with dependency errors; going to call this "not my fault"

chilts:
Travis is now a happy bunny. Tests pass locally. Code LGTM. r+.

In other news, when Travis failed in the past I would just click 'restart', but if just one variation of the tests fail you can restart just that one! Other people may well have already known this, but just putting it here in case. :)

PR #122, opened 2015-06-24
fix(display_name): allow a blank display name

zaach:
We should probably allow users to clear their display name.

@vladikoff r?

### DIFF
--- a/test/api.js
+++ b/test/api.js
@@ -642,6 +642,29 @@ describe('/display_name', function() {
         assert.equal(res.displayName, NAME);
       });
     });
+
+    it('should allow an empty display name', function() {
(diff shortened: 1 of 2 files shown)


test/api.js
```diff
@@ -642,6 +642,29 @@ describe('/display_name', function() {
         assert.equal(res.displayName, NAME);
       });
     });
+
+    it('should post an empty display name', function() {
```
rfk: nit: "should allow" would be more descriptive than "should post"

rfk:
r+ modulo nits

vladikoff:
Sorry @zaach lost this PR,  thought this was merged already, but yeah r+

PR #129, opened 2015-07-28
feat(events): add events to delete user data when account is deleted

### LINKED ISSUE #127: Delete profile data when a user deletes their account
Similarly to sync, oauth and other FxA-attached services, we should delete any data stored on behalf of a user if they delete their FxA account.  I think we can safely copy the approach taken in oauth-server as a starting point; we have elaborate plans for a system to unify this, but it's on the backburner for now.

seanmonstar:
Closes #127

A missing piece is tests for this... I don't know how specifically to test the `Sink`...


lib/events.js
```diff
@@ -0,0 +1,70 @@
...
+    if (message.event === 'delete') {
+      var userId = message.uid.split('@')[0];
+      if (!HEX_STRING.test(userId)) {
+        message.del();
+        return logger.warn('badDelete', { userId: userId });
+      }
+      P.all([
+        db.getAvatars(userId).then(function(avatars) {
+          return avatars.map(function(avatar) {
+            return workers.delete(avatar.id).then(function() {
```
rfk: If the avatar no longer exists on disk, will `workers.delete` fail or succeed?  If it errors out, ISTM we could get into a situation where the avatar is still referenced in the db but doesn't exist on disk, meaning that the cleanup task can't suceed no matter how many times it gets retried.

rfk:
>  I don't know how specifically to test the Sink... 

Per IRL discussion, let's factor the processing logic out into a function that can be tested in isolation, and assume that the Sink class itself works as advertised.  We could also consider bringing over sinon and the mocking stuff that @philbooth recently added in fxa-oauth-server, if necessary.

ckolos:
This actually requires some work on the ops side; the current set of permissions on the bucket will not allow this to occur.

seanmonstar:
@ckolos oh ok. should i file a bug/issue so we can track, and link it here?

ckolos:
@seanmonstar - please

ckolos:
@jrgm https://github.com/mozilla-services/svcops/pull/615

lib/events.js
seanmonstar: According to http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#deleteObject-property it won't error if there's no object.

seanmonstar:
@rfkelly so, I've added tests for the `onMessage` part....

bin/server.js
```diff
@@ -3,10 +3,17 @@
...
 logger.debug('config', config);
-server.start(function() {
-  logger.info('listening', server.info.uri);
+db.ping().done(function() {
+  server.start(function() {
+    logger.info('listening', server.info.uri);
+  });
+  events.start();
+}, function(err) {
+  logger.critical('db.ping', err);
```
rfk: Should we do something here to exist with a non-zero exit code, or will that will happen automagically despite handling the failure?

lib/workers.js
```diff
@@ -0,0 +1,62 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const P = require('./promise');
+
+const AppError = require('./error');
+const config = require('./config');
+const logger = require('./logging')('workers');
```
rfk: Naming nit: I feel like "workers" is too generic a name here.  Unless we expect this to grow responsibilities beyond handling avatars, perhaps we should name it something more specific, so that calling code reads like `avatarHandler.upload() or maybe `workers.uploadAvatar()` rather than just `workers.upload()` which doesn't provide much information.

test/events.js
```diff
@@ -0,0 +1,95 @@
...
+    function Message(del) {
+      return {
+        event: 'delete',
+        uid: UID + '@accounts.firefox.com',
+        del: del
+      };
+    }
+
+    before(function() {
+      this.slow(2000);
```
rfk: Out of curiosity, why so slow?

test/events.js
```diff
@@ -0,0 +1,95 @@
...
+      db.setDisplayName(UID, 'foo bar').then(function() {
+        events.onData(new Message(function() {
+          db.getDisplayName(UID).then(function(profile) {
+            assert.equal(profile.displayName, undefined);
+          }).done(done, done);
+        }));
+      });
+    });
+  });
+});
```
rfk: Please add some edge-case-handling tests, in particular:
* non-"delete" messages are discarded
* errors during deletion result in the message staying in the queue for eventual retry

rfk:
Also travis bustage, but looks like it might have been a dependency/install issue?

test/events.js
seanmonstar: Ah, I copied this from the api test. Uploading was the first thing I did in this repo, and the tests used to be slow, until I added the `mock` stuff. I must have forgotten to take these out. Killed.

lib/workers.js
seanmonstar: I already have `img`, which does the actual storing of the avatars. I needed something else that explained 'this uses the worker queue to do the action!'.

rfk: Perhaps `imgworker` or similar?  Naming is hard.

test/events.js
```diff
@@ -0,0 +1,123 @@
...
+      });
+
+    });
+
+    it('should ignore unknown messages', function(done) {
+      db.setDisplayName(UID, 'foo bar').then(function() {
+        events.onData(new Message('baz', function() {
+          db.getDisplayName(UID).then(function(profile) {
+            assert.equal(profile.displayName, 'foo bar');
+          }).done(done, done);
```
rfk: Thanks.  Do we have any easy way to assert that the `del` callback was actually executed on this message?  I guess we could just flip a flag variable inside the callback and then check it after promise fulfills.

rfk: Oh, onData doesn't return a promise does it...so I guess the test would just time out in this case?

test/events.js
```diff
@@ -0,0 +1,123 @@
...
+          assert(false, 'message.del() should not be called');
+        }));
+        mock.log('events', function(record) {
+          return record.levelname === 'ERROR'
+              && record.args[0] === 'removeUser';
+        });
+
+        setTimeout(function() {
+          done();
+        }, 50);
```
rfk: Could we make `onData` return a promise to avoid using `setTimeout` here, or would it get too gnarly too quickly?

test/events.js
seanmonstar: Correct, the test would timeout if the `done` method isn't called.

test/events.js
seanmonstar: `onData` is set as the event handler, and nodejs inspects the return value of those. Also, with Promises, you can either return them, saying "it's someone else's responsibility now", or you can call `p.done()`, which will throw an error if it hasn't been previously handled. I'd rather call `done` inside that event handler, than hoping somewhere else does.

However, I can move the timeout into the log mock, since that's a requirement for the test to pass, and happens at the end.

seanmonstar:
Updated.

lib/img-workers.js
```diff
@@ -0,0 +1,62 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+const P = require('./promise');
+
+const AppError = require('./error');
+const config = require('./config');
+const logger = require('./logging')('img_workers');
```
rfk: Thanks for the rename, I think it reads clearer now.  IIUC this will change the name of some log messages from "workers" to "img" workers, so we should check with @jrgm whether any monitoring on the ops side will need to change to match.  @jrgm?

lib/events.js
```diff
@@ -0,0 +1,66 @@
...
+      db.getAvatars(userId).then(function(avatars) {
+        return P.all(avatars.map(function(avatar) {
+          return workers.delete(avatar.id).then(function() {
+            return db.deleteAvatar(avatar.id);
+          });
+        }));
+      }),
+      db.removeProfile(userId)
+    ])
+      .done(function () {
```
rfk: maybe nit: something about the indentation here doesn't seem quite right.

We seem to favour `...]).done(...` on a sigle line in this repo, but maybe that doesn't work here because we've got two separate callbacks defined as inline functions?  I don't have a specific alternate suggestion, just wanted to flag it as something that was odd enough to be noticed.

rfk:
@seanmonstar I think you need to regenerate npm-shrinkwrap to get travis working with the new dependencies.  r+ pending a green from travis and a comment from @jrgm on the logging name change.

lib/img-workers.js
seanmonstar: This file did not exist prior to this PR, so it will likely be changes from `routes.avatar.upload...` to `img_workers.upload`.

rfk:
Travis is looking good, assigning to @jrgm just to sanity-check the logging name change per https://github.com/mozilla/fxa-profile-server/pull/129#discussion_r37151407.

FWIW I can't see any instances of 'avatar.upload' in puppet-config

jrgm:
Yeah, I don't see a problem with that name change.

PR #136, opened 2015-08-20
fix(config): add options events.region and events.queueUrl

jrgm:
I assume you meant to add these options. Merging when travis passes because latest.dev.lcip.org is currently broken.


rfk:
Ugh, r? fail on my part in #129, sorry @jrgm.  This LGTM.

PR #137, opened 2015-08-20
fix(display_name): Don't allow control characters in the display_name field

### LINKED ISSUE #126: Limit allowed characters in display-name field
We don't want to accidentally XSS ourselves like @pdehaan did over in https://github.com/mozilla/fxa-content-server/issues/2794.

We should limit the allows characters in display names to alphnumeric plus some sensible metachars.

rfk:
Fixes #126.

I went with a very slim list of control characters that IMHO have no business being in a display name, and would only be there to cause trouble.  I also added tests that various other unicodey things are accepted without trouble, which seemed to work except for one issue with JSON.parse-ing the result payload in the tests.

@seanmonstar r?

### DIFF
--- a/test/api.js
+++ b/test/api.js
@@ -603,7 +603,7 @@ describe('/display_name', function() {
       });
     });
 
-    it('should NOT return a profile if wrong scope', function() {
+    it('should NOT return a display_name if wrong scope', function() {
       mock.token({
         user: USERID,
         email: 'user@example.domain',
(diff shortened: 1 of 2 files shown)


test/api.js
```diff
@@ -661,9 +661,109 @@ describe('/display_name', function() {
...
+        });
+      }).then(function(res) {
+        assert.equal(res.statusCode, 404);
+      });
+    });
+
+    it('should accept a variety of unicode characters', function() {
+      var NAMES = [
+        'André Citroën',
+        'the unblinking ಠ_ಠ of ckarlof',
```
seanmonstar: :+1:

PR #142, opened 2015-09-07
fix(version): use explicit path with git-config

jrgm:
Apologies for futzing with this. 

I realized that git-config should use an explicit path to the config location (because, on fxa-dev boxes, fxa-oauth-server starts from `/`, not the checked out repo, and so can't find the config). 

Also, eliminated and unused IFFE, and added a newline to the output because nicer.


rfk:
>  because nicer.

:+1:

Issue #146, opened 2015-09-16
"avatar.get" activityEvent


vladikoff:
Hey @rfk, we are looking at add `"avatar.used"` (name pending) activityEvent that would originate from the oauth / profile server and then calculate retention for users that have "avatar.used".  This might be an oauth server issue though. 

cc @philbooth

vladikoff:
@rfk we want to add it sooner than later so we can start gathering data

rfk:
This repo is probably the right place, adding it to one or more of the routes in https://github.com/mozilla/fxa-profile-server/tree/master/lib/routes/avatar

PR #147, opened 2015-09-18
feat(logging): add avatar.get activity event

vladikoff:


vladikoff:
Fixes https://github.com/mozilla/fxa-profile-server/issues/146

lib/routes/avatar/get.js
```diff
@@ -33,13 +34,19 @@ module.exports = {
...
+    db.getSelectedAvatar(uid)
       .then(avatarOrEmpty)
       .done(function (result) {
         var rep = reply(result);
         if (result.id) {
           rep = rep.etag(result.id);
         }
+        var info = {
+          event: 'avatar.get',
+          uid: req.uid
```
philbooth: Should this just be `uid`?

vladikoff: :+1:

lib/routes/avatar/get.js
```diff
@@ -33,13 +34,19 @@ module.exports = {
...
       .done(function (result) {
         var rep = reply(result);
         if (result.id) {
           rep = rep.etag(result.id);
         }
+        var info = {
+          event: 'avatar.get',
+          uid: uid
+        };
+        logger.info('activityEvent', info);
```
rfk: IIUC this will log an event whenever something *attempts* to read the user's avatar, even if they don't actually have an avatar set (hence the use of `avatarOrEmpty` above).  We could move it into the `if (result.id)` block to ensure it only emits if there's actually an avatar to read.

@vladikoff is it better to measure avatar-reads or avatar-writes for our metrics purposes?

vladikoff: > @vladikoff is it better to measure avatar-reads or avatar-writes for our metrics purposes?

I think avatar-write should be more useful, but avatar-read also has a use case:
with 
avatar-write:
"Users who uploaded an avatar and still using" vs "Users who did not and still using"

avatar-get:
"User fetching avatar after X weeks"

vladikoff:
## Works:
```
 NODE_ENV=test mocha test/api.js --grep="avatar.post"   
```
## Not Works:
```
npm test
```

vladikoff:
@rfk @seanmonstar r?

vladikoff:
We need this one before we cut v0.46.0

vladikoff:
ping @seanmonstar

PR #150, opened 2015-10-01
Return 204 if user does not have a display name set

### LINKED ISSUE #144: /v1/display_name returns 404 if not set
via irc:
  > jrgm> yeah, so /v1/display_name returns 404 for no name set. What if that was 200 OK with the empty string instead.

seanmonstar:
Closes #144 

r? @zaach

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -292,6 +292,8 @@ curl -v \
 }
 ```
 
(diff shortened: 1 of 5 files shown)


zaach:
:+1:

PR #154, opened 2015-10-12
fix(server): prevent null exception when oauth server is down

### LINKED ISSUE #151: prod 500 errors for /v1/profile on possible network glitch
This may have just been a network glitch but there was a burst of 500's in production, about 24. But this code could use a null check. - https://github.com/mozilla/fxa-profile-server/blob/master/lib/server/web.js#L64

```
{"Timestamp":1443753325303000000,"Logger":"fxa-profile-server","Type":"server.summary","Severity":2,"Pid":2069,"EnvVersion":"2.0",
 "Fields":{"code":500,"method":"get","errno":999,"path":"/v1/profile",
  "agent":"Mozilla/5.0 (Windows NT 10.0; WOW64; rv:41.0) Gecko/20100101 Firefox/41.0","t":22156,
  "stack":"TypeError: Uncaught error: Cannot read property 'code' of undefined
    at Request._callback (/data/fxa-profile-server/lib/server/web.js:64:19)
    at Request.self.callback (/data/fxa-profile-server/node_modules/request/request.js:372:22)
    at Request.emit (events.js:98:17)
    at Request.<anonymous> (/data/fxa-profile-server/node_modules/request/request.js:1310:14)
    at Request.emit (events.js:117:20)
    at IncomingMessage.<anonymous> (/data/fxa-profile-server/node_modules/request/request.js:1258:12)
    at IncomingMessage.emit (events.js:117:20)
    at _stream_readable.js:944:16
    at process._tickDomainCallback (node.js:492:13)"}}
```

seanmonstar:
Closes #151

@jrgm r?

### DIFF
--- a/lib/server/web.js
+++ b/lib/server/web.js
@@ -57,8 +57,8 @@ exports.create = function createServer() {
             token: token
           }
         }, function(err, resp, body) {
-          if (err) {
-            logger.error('auth', err);
+          if (err || resp.statusCode >= 500) {
+            err = err || resp.statusMessage || 'unknown';
             return reply(AppError.oauthError(err));
           }
           if (body.code >= 400) {
--- a/test/api.js
+++ b/test/api.js
@@ -109,6 +109,26 @@ describe('/profile', function() {
     });
   });
 
+  it('should handle oauth server failure', function() {
+    mock.tokenFailure();
+
+    mock.log('server', function(rec) {
+      return rec.levelname === 'ERROR'
+          && rec.args[0] === 'summary'
+          && rec.args[1].path === '/v1/profile';
+    });
+
+    return Server.api.get({
+      url: '/profile',
+      headers: {
+        authorization: 'Bearer ' + tok
+      }
+    }).then(function(res) {
+      assert.equal(res.statusCode, 503);
+      assert.equal(res.result.errno, 104);
+    });
+  });
+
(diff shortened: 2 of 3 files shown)


lib/server/web.js
```diff
@@ -57,8 +57,8 @@ exports.create = function createServer() {
             token: token
           }
         }, function(err, resp, body) {
-          if (err) {
-            logger.error('auth', err);
+          if (err || resp.statusCode >= 500) {
+            err = err || resp.statusMessage || 'unknown';
             return reply(AppError.oauthError(err));
```
jrgm: So, I'm not sure that during the failure period that resp or resp.statusCode had a value. I don't know why `err` was not truthy though. Hand-waving, but it might need to listen for an `error` event to detect the failure. 

But perhaps just doing 'if !body then return error()' is what is needed.

seanmonstar: It's because I was checking `body`, which I assumed was a json payload. If `err` is falsey, then `resp` **will** be a `http.IncomingMessage`. Just because we received a response doesn't mean the response had a body (the part after the headers), or that it was of content-type: json.

In this case, it was likely just `HTTP/1.1 500 Internal Server Error\r\nServer: ...\r\nDate: ...\r\n\r\n`, and so `body` was null.

seanmonstar:
@rfk r?

PR #166, opened 2015-11-09
feat(email): fetch email from auth-server /account/profile

### LINKED ISSUE #165: Fetch the email from the auth-server instead of the oauth server

seanmonstar:
Closes #165

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -14,6 +14,14 @@ const conf = convict({
       default: 1
     }
   },
+  authServer: {
+    url: {
+      doc: 'URL of fxa-auth-server',
+      env: 'AUTH_SERVER_URL',
+      format: 'url',
+      default: 'http://127.0.0.1:9000/v1'
+    }
+  },
   db: {
     driver: {
       env: 'DB',
--- a/lib/routes/email.js
+++ b/lib/routes/email.js
@@ -5,6 +5,11 @@
 const Joi = require('joi');
 
 const AppError = require('../error');
+const config = require('../config');
+const logger = require('../logging')('routes.email');
+const request = require('../request');
+
+const AUTH_SERVER_URL = config.get('authServer.url') + '/account/profile';
 
 module.exports = {
   auth: {
@@ -17,14 +22,30 @@ module.exports = {
     }
   },
   handler: function email(req, reply) {
-    var email = req.auth.credentials.email;
-    if (email) {
+    request.get(AUTH_SERVER_URL, {
+      headers: {
+        Authorization: 'Bearer ' + req.auth.token
+      },
+      json: true
+    }, function(err, res, body) {
+      if (err) {
+        logger.error('request.auth_server.network', err);
+        return reply(new AppError.authError('network error'));
+      }
+      if (res.statusCode >= 400) {
+        logger.error('request.auth_server.fail', { code: res.statusCode });
+        return reply(new AppError.authError('auth server error'));
+      }
+
+      if (!body || !body.email) {
+        return reply(
+          new AppError.authError('email field missing from auth response')
+        );
+      }
       reply({
-        email: email
+        email: body.email
       });
-    } else {
-      reply(new AppError.oauthError('email field missing from oauth response'));
-    }
+    });
   }
 };
 
--- a/lib/server/web.js
+++ b/lib/server/web.js
@@ -54,11 +54,13 @@ exports.create = function createServer() {
         request.post({
           url: url,
           json: {
-            token: token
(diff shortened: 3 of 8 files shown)


lib/config.js
```diff
@@ -14,6 +14,13 @@ const conf = convict({
       default: 1
     }
   },
+  authServer: {
+    url: {
+      doc: 'URL of fxa-auth-server',
+      format: 'url',
```
vladikoff: I think we are trying to have ENV variables for these. seems like this one is missing a way to set this setting via an ENV var.

Ref: https://github.com/mozilla/fxa-auth-server/blob/master/config/index.js#L85-L102

seanmonstar:
I've also updated this PR to pass `email=false` to the oauth server, which means this depends on https://github.com/mozilla/fxa-oauth-server/pull/359

lib/config.js
```diff
@@ -14,6 +14,14 @@ const conf = convict({
...
     }
   },
+  authServer: {
+    url: {
+      doc: 'URL of fxa-auth-server',
+      env: 'AUTH_SERVER_URL',
+      format: 'url',
+      default: 'http://127.0.0.1:9000/v1'
+    }
+  },
```
rfk: Style nit: we should put this next to the `oauth.url` config option and ensure they're named consistently.  Right now it's "oauth.url" and "authServer.url", and I don't strongly prefer one over the other, but let's pick a consistent style.

rfk: Also I'll be super glad when we smoosh these things together into a single endpoint...

lib/routes/email.js
```diff
@@ -17,14 +22,25 @@ module.exports = {
     }
   },
   handler: function email(req, reply) {
-    var email = req.auth.credentials.email;
-    if (email) {
+    request.get(AUTH_SERVER_URL, {
+      headers: {
+        Authorization: 'Bearer ' + req.auth.token
+      },
+      json: true
```
rfk: Will this cause the oauth server to have to verity the token twice, once for the profile-server's own scope-checking, then again for the auth-server's permission checks?

If so, I'm OK with it while we're in the process of restructuring things, but we should flag it with @jrgm and @ckolos because they'll see a spike in oauth server traffic.

lib/routes/email.js
```diff
@@ -17,14 +22,25 @@ module.exports = {
...
+    }, function(err, res, body) {
+      if (err) {
+        logger.error('request.auth_server.network', err);
+        return reply(new AppError.authError('network error'));
+      }
+      if (res.statusCode >= 400) {
+        logger.error('request.auth_server.fail', { code: res.statusCode });
+        return reply(new AppError.authError('auth server error'));
```
rfk: I wish we could reflect through the error number form the auth-server here, but I guess that doesn't make much sense since auth and profile API errornos are different.

lib/routes/email.js
```diff
@@ -17,14 +22,25 @@ module.exports = {
...
+      }
+
       reply({
-        email: email
+        email: body.email
       });
-    } else {
-      reply(new AppError.oauthError('email field missing from oauth response'));
-    }
```
rfk: Should we keep this "if there wasn't an `email` field then throw an error" logic just in case?

lib/server/web.js
```diff
@@ -54,11 +54,13 @@ exports.create = function createServer() {
         request.post({
           url: url,
           json: {
-            token: token
+            token: token,
+            email: false
```
rfk: It may be worth a quick comment inline about what this field's doing on here.

rfk:
I left a couple of nits and q's, but this looks solid to me :+1: 

Needs a rebase, which will hopefully fix the travis bustage.

lib/config.js
seanmonstar: I wanted to follow style, but having a config value `auth` sounds quite different than `authServer`.

lib/routes/email.js
seanmonstar: Yes, it will. There's no real way around it as long as the auth server has the email, and the profile server has other details.

lib/config.js
seanmonstar: As for position in the config file, I've been trying to keep it alphabetically sorted.

rfk: > I've been trying to keep it alphabetically sorted

:+1:

lib/routes/email.js
seanmonstar: Done.

vladikoff:
This broke local development and latest. Reverting this PR

PR #170, opened 2015-11-11
fix(config): adjust lib gm limits for aws

### LINKED ISSUE #167: limits code is resulting in only the '_large' variant being stored in s3
re: https://github.com/mozilla/fxa-profile-server/pull/157/files#diff-e1b05bbd72783375d7e1f2df53824f9fR25

In stage with train-49, only the '_large' variant is being stored in s3.

vladikoff:
Fixes #167


vladikoff:
@seanmonstar r?

PR #172, opened 2015-11-13
fix(upload): add gm image identification

### LINKED ISSUE #96: POST /avatar/upload should validate image payload
If the body of the request doesn't look like image data (e.g. `{}`) the request should fail.

vladikoff:
Fixes #96


lib/img-workers.js
```diff
@@ -30,13 +30,17 @@ exports.upload = function upload(id, payload, headers) {
     };
     logger.verbose('upload', url);
     payload.pipe(request.post(url, opts, function(err, res, body) {
+      var nestedError = null;
+      if (body && body[0] && body[0].error) {
```
vladikoff: from my logs this returned an array with an error  and not `body.error`. I wonder if it ever returns body.error?

lib/compute/image-cc.js
```diff
@@ -23,15 +23,23 @@ function processImage(src, width, height) {
...
         if (err) {
-          reject(err);
-        } else {
-          resolve(buf);
+          // if gm cannot identify this image, then reject
+          return reject (err);
         }
-      });
+
+        gm(src).resize(width, height, '>')
```
seanmonstar: should use the `limit` function again here, no?

seanmonstar:
r=me with comment addressed.

PR #177, opened 2015-11-26
fix(email): handle no-such-account errors when fetching email from auth server

### LINKED ISSUE #176: prod errors train-50
Example: 

```
{"Timestamp":1448489854219000000,"Logger":"fxa-profile-server","Type":"routes.email.request.auth_server.fail","Severity":2,"Pid":2054,"EnvVersion":"2.0","Fields":{"code":400}}
{"Timestamp":1448489854221000000,"Logger":"fxa-profile-server","Type":"server.summary","Severity":2,"Pid":2054,"EnvVersion":"2.0","Fields":{"code":503,"method":"get","errno":105,"path":"/v1/email","agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.11; rv:42.0) Gecko/20100101 Firefox/42.0","t":27,"stack":"Error: Auth server error
    at new authError (/data/fxa-profile-server/lib/error.js:142:10)
    at Request._callback (/data/fxa-profile-server/lib/routes/email.js:37:22)
    at Request.self.callback (/data/fxa-profile-server/node_modules/request/request.js:372:22)
    at Request.emit (events.js:98:17)
    at Request.<anonymous> (/data/fxa-profile-server/node_modules/request/request.js:1310:14)
    at Request.emit (events.js:117:20)
    at IncomingMessage.<anonymous> (/data/fxa-profile-server/node_modules/request/request.js:1258:12)
    at IncomingMessage.emit (events.js:117:20)
    at _stream_readable.js:944:16
    at process._tickDomainCallback (node.js:492:13)"}}
{"Timestamp":1448489854223000000,"Logger":"fxa-profile-server","Type":"batch.email.503","Severity":2,"Pid":2054,"EnvVersion":"2.0","Fields":{"code":503,"errno":105,"error":"Service Unavailable","message":"Auth server error","cause":"auth server error"}}
{"Timestamp":1448489854224000000,"Logger":"fxa-profile-server","Type":"server.summary","Severity":2,"Pid":2054,"EnvVersion":"2.0","Fields":{"code":503,"method":"get","errno":105,"path":"/v1/profile","agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.11; rv:42.0) Gecko/20100101 Firefox/42.0","t":86,"stack":"Error: Auth server error
    at Function.from (/data/fxa-profile-server/lib/error.js:51:13)
    at /data/fxa-profile-server/lib/batch.js:47:26
    at tryCatch1 (/data/fxa-profile-server/node_modules/bluebird/js/main/util.js:45:21)
    at Promise$_callHandler [as _callHandler] (/data/fxa-profile-server/node_modules/bluebird/js/main/promise.js:660:13)
    at Promise$_settlePromiseFromHandler [as _settlePromiseFromHandler] (/data/fxa-profile-server/node_modules/bluebird/js/main/promise.js:675:18)
    at Promise$_settlePromiseAt [as _settlePromiseAt] (/data/fxa-profile-server/node_modules/bluebird/js/main/promise.js:845:14)
    at Promise$_settlePromises [as _settlePromises] (/data/fxa-profile-server/node_modules/bluebird/js/main/promise.js:988:14)
    at Promise.b (domain.js:183:18)
    at Async$_consumeFunctionBuffer [as _consumeFunctionBuffer] (/data/fxa-profile-server/node_modules/bluebird/js/main/async.js:77:12)
    at Async$consumeFunctionBuffer (/data/fxa-profile-server/node_modules/bluebird/js/main/async.js:40:14)
    at process._tickDomainCallback (node.js:492:13)"}}
```

rfk:
Fixes #176 

Assuming @seanmonstar isn't working for the rest of the week, @vladikoff r?


rfk:
Per https://github.com/mozilla/fxa-profile-server/issues/176#issuecomment-159753932, the problem here is that an account can be deleted on the auth-server, but still have oauth tokens that are valid and clients that are trying to use them.  They will *eventually* be deleted by our background cleanup jobs in the oauth server, but there's a weirdness window until they are.  Distributed state, wheeee!

vladikoff:
Travis blocked by
```
  
Running "eslint:app" (eslint) task
lib/routes/email.js
  36:12  error  body is already defined  no-redeclare
  36:12  error  body is already defined  no-redeclare
  36:12  error  body is already defined  no-redeclare
  36:12  error  body is already defined  no-redeclare
✖ 4 problems (4 errors, 0 warnings)
Warning: Task "eslint:app" failed. Use --force to continue.
```

lib/routes/email.js
```diff
@@ -33,7 +33,15 @@ module.exports = {
...
+        body = body && body.code ? body : { code: res.statusCode };
+        // It's possible for the account to be deleted on the auth-server
+        // but its oauth tokens are still valid, pending cleanup.
+        // We just treat it as if auth failed.
+        if (body.errno === 102) {
+          logger.info('request.auth_server.fail', body);
+          return reply(new AppError.unauthorized(body.message));
+        }
+        logger.error('request.auth_server.fail', body);
         return reply(new AppError.authError('auth server error'));
```
jrgm: So, this seems wrong even with the change. Any 400 response, except one with subcode 102, is being logged and reported back to the client as a '503 Service Unavailable'. There's no service that is unavailable if a 4xx came back from the auth service.

rfk: The other possible 4XXs from that endpoint are:

* a `401` if the bearer token is invalid; this should be passed directly back to the client
* various `400` errors due to malformed requests

Seems we're better off turning them into 400s rather than 503s, I'll update the PR.

rfk:
@jrgm this version should report all 4XX errors from auth-server back as a 4XX error from the profile-server, without mixing up the (unfortunately differing) errno values between the two servers.

jrgm:
@rfk thanks. This looks good to me.

jrgm:
Actually, not the right place, but I'll note this here anyways. In addition to the issue with 5xx errors, two other changes in metrics occurred when this was live. 

1) latency for /v1/profile increased by about 60ms average (which makes sense since it now takes a longer series of calls to complete).
2) /verify traffic on oauth-server doubled. Not sure if this is entirely expected. Comments?

rfk:
Re-assigning this to @seanmonstar for comment on the above two points, and a final r?

seanmonstar:
The reason I decided to make 400 status codes 500s was because the bearer token would have already been sent to `/v1/verify` before getting to this code, and hapi will have enforced the `profile:email` scope, so most 400s would mean something fundamentally went wrong.

----

@jrgm 

1. Yep, profile server now requires a second request, to the auth server, to get the email.
2. Also expected. The profile server must verify the token first, to allow all other scopes it accepts, and then once it asks the auth server for the email, the auth server also verifies the token. Unfortunate double dipping, but can be optimized by sharing read access to the auth server db.

rfk:
> so most 400s would mean something fundamentally went wrong

*nod*; fwiw I don't feel strongly either way, would be happy to make the "all other 400 errors" case either a 400 or a 500.

seanmonstar:
Other than "account deleted, still in oauth sqs queue", every other 400 should have been hit previously in the profile server, right? So, if we *did* see that 400, I'd want the monitoring to be like "yo wtf, things be 500ing" instead of "eh, someone failed to login, carry on".

rfk:
>  if we did see that 400, I'd want the monitoring to be like "yo wtf, things be 500ing" instead of 
> "eh, someone failed to login, carry on".

@jrgm thoughts?

jrgm:
So, this still doesn't strike me as a server error (at least not in the same disaster category as "cannot connect to database", exceptions, network errrors). But if there is **no way** that a 5xx error would ever result (aside from the now-handled 400/102, then I guess it's moot).

rfk:
IIUC any other 400s we get back from the auth-server will be *our fault*, e.g. the profile-server sent some malformed request to the auth-server.  Reporting those as 500s seems ok since there's nothing that the client did wrong.

rfk:
Based on above discussion, I've changed the handling of unrecognized 400s from the auth-server to be a "500 Internal Server Error", distinct from the "503 Service Unavailable" that we'll get if the auth-server is down in some way.  This seems to capture the fact that it was our fault, not the clients, while not conflating it with operational service outages.

@seanmonstar back to you for final r?, and if you're happy, a patch release tag.

rfk:
Travis sees to be having a problem, but these tests pass locally for me...

PR #179, opened 2015-12-15
feat(openid): make /v1/profile act as the OIDC UserInfo endpoint

### LINKED ISSUE #175: Implement openid-connect-compatible "UserInfo" endpoint
The OpenID Connect spec defines a "UserInfo" endpoint, through which a relier can fetch additional information about the user [1].  We should implement it.

[1] http://openid.net/specs/openid-connect-core-1_0.html#UserInfo

seanmonstar:
Closes #175

### DIFF
--- a/lib/routes/email.js
+++ b/lib/routes/email.js
@@ -14,7 +14,7 @@ const AUTH_SERVER_URL = config.get('authServer.url') + '/account/profile';
 module.exports = {
   auth: {
     strategy: 'oauth',
-    scope: ['profile', 'profile:email']
+    scope: ['profile', 'profile:email', /* openid-connect scope */'email' ]
   },
   response: {
     schema: {
--- a/lib/routes/profile.js
+++ b/lib/routes/profile.js
@@ -8,11 +8,12 @@ const checksum = require('checksum');
 
 const batch = require('../batch');
 
-function hasProfileScope(scopes) {
+function hasAllowedScope(scopes) {
   for (var i = 0, len = scopes.length; i < len; i++) {
     var scope = scopes[i];
     // careful to not match a scope of 'profilebogie'
-    if (scope === 'profile' || scope.indexOf('profile:') === 0) {
+    if (scope === 'profile' || scope === 'email'
+        || scope.indexOf('profile:') === 0) {
       return true;
     }
   }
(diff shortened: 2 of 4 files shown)


seanmonstar:
It may make sense to eventual create a `/v1/userinfo` endpoint that supports the proper semantics of the various scopes defined for OpenID Connect, but doing so would only require updating the openid-configuration well-known file to point at the new endpoint, and all would be well.

lib/routes/profile.js
```diff
@@ -8,11 +8,19 @@ const checksum = require('checksum');
...
-function hasProfileScope(scopes) {
+const ALLOWED_SCOPES = {
+  'profile': true,
+  'profile:': true,
+  //openid connect
+  'email': true
+};
+
+function hasAllowedScope(scopes) {
+  console.log('hasAllowedScope: ', scopes);
```
rfk: console.log

lib/routes/profile.js
```diff
@@ -8,11 +8,19 @@ const checksum = require('checksum');
...
   for (var i = 0, len = scopes.length; i < len; i++) {
-    var scope = scopes[i];
+    var scope = scopes[i].substr(0, 'profile:'.length);
```
rfk: Neat trick, but it seems a little implicit and fragile to me, e.g. it depends on `'profile:'` being the longest entry in the `ALLOWED_SCOPES` set.  Given how important scope-matching is to security, I think we should prefer something simple and verbose over something short and clever.

(Of course, what we should really do is use a common helper lib for doing scope matching so that we only have to get it right once, but that's outside the scope of this bug).

lib/routes/email.js
```diff
@@ -14,7 +14,7 @@ const AUTH_SERVER_URL = config.get('authServer.url') + '/account/profile';
 module.exports = {
   auth: {
     strategy: 'oauth',
-    scope: ['profile', 'profile:email']
+    scope: ['profile', 'profile:email', /* openid-connect scope */'email' ]
```
rfk: I suspect we'll have to also add this logic in the auth-server `/account/profile` endpoint, otherwise we won't be able to read it out in a non-mocked setting.

rfk:
> It may make sense to eventual create a /v1/userinfo endpoint that supports 
> the proper semantics of the various scopes defined for OpenID Connect

Agreed, this seems like a good place to start and it's simple for us to continue refining it later.

rfk:
Also, deps-related travis bustage which I hope is just transient noise

lib/routes/email.js
seanmonstar: Ah yes. How irritating.

lib/routes/profile.js
seanmonstar: Ok, returned to explicit `===` and `indexOf`.

rfk:
LGTM, but I don't understand why travis is complaining.

seanmonstar:
Er... installing graphicsmagick is failing. I see at the top of the travis build, a link to this: https://blog.travis-ci.com/2015-11-27-moving-to-a-more-elastic-future

vladikoff:
Maybe try this: https://github.com/mozilla/fxa-local-dev/blob/master/.travis.yml#L14 @seanmonstar

PR #180, opened 2015-12-17
chore(travis): update to use nosudo travis infra

seanmonstar:


.travis.yml
```diff
@@ -21,9 +27,7 @@ env:
   - NODE_ENV=test DB=mysql IMG=aws
 
 before_install:
-  - npm install -g npm@2
```
jrgm: Actually, node 0.10 ships with with npm @ 1.4.x, which can fail to installl modules correctly at times. (Yeah, overwriting the npm@2 that comes with 0.12 and 4.x is lame, but npm@1 is lamer).

seanmonstar: I believe overriding it globally requires sudo, so I deleted this line..

jrgm: travis-ci sets the node version with `nvm` so `npm` is installed as the travis user under `/home/travis/.nvm` so overwriting does not require sudo - e.g., https://travis-ci.org/mozilla/fxa-auth-server/jobs/97399693#L256

PR #185, opened 2016-02-15
chore(deps): Update all the dependencies

### LINKED ISSUE #184: upgrade mozlog to 2.0.3
Picks up this fix for exception logging - mozilla/mozlog@e11def1 (of course).

But in updating package.json, removing npm-shrinkwrap.json and `npm install`, `npm test` barfs first due to (I think) the blanket issue from https://github.com/mozilla/fxa-oauth-server/pull/380.

However, pinning that to blanket@1.1.6, I still have a bunch of test failures, which I assume is due to one or more of the (nominally) compatible bumps in semver for dependencies using `~` or `^`.

Punting!

rfk:
Fixes #184, and since it had breakage anyway, I sent ahead and updated all the other dependencies as well.  I also removed the `~` and `^` version specifiers in favor of pinning all deps to a specific version, since AFAICT our use of shrinkwrap means they're not that useful.

The most interesting thing here is the Hapi 8, which is not the latest upstream but is at least in line with what fxa-auth-server is running.  This had some breaking API changes to deal with, but nothing too tricky.  Getting all the way to latest Hapi means switching to node 4.

@seanmonstar r?


seanmonstar:
:eyes: :+1:

PR #189, opened 2016-05-18
feat(docker): Add CloudOps Dockerfile & CircleCI build instructions

jbuck:


seanmonstar:
It looks fine to me, but I don't know anything about docker or circleci, or a way to test these. Should I just merge, or is there someone else who would know?

Dockerfile-build
```diff
@@ -0,0 +1,22 @@
...
+RUN groupadd --gid 10001 app && \
+    useradd --uid 10001 --gid 10001 --home /app --create-home app
+
+RUN apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get -y install graphicsmagick && \
+    apt-get clean
+
+USER app
+
+COPY . /app
```
mostlygeek: 1. you should copy ./package.json /app/package.json first 
1. then run `npm install`
1. then copy . /app

This way you get better usage out of the docker image cache if dependencies don't change

circle.yml
```diff
@@ -0,0 +1,44 @@
...
+# DOCKER_PASS
+#
+machine:
+  services:
+    - docker
+dependencies:
+  override:
+    - docker info
+
+    - printf '{"commit":"%s","version":"%s","source":"https://github.com/%s/%s","build":"%s"}\n' "$CIRCLE_SHA1" "$CIRCLE_TAG" "$CIRCLE_PROJECT_USERNAME" "$CIRCLE_PROJECT_REPONAME" "$CIRCLE_BUILD_URL" > config/version.json
```
mostlygeek: would be nicer if you split the lines like this: 
```yaml
    - >
        printf '{"commit":"%s","version":"%s","source":"https://github.com/%s/%s","build":"%s"}\n' 
        "$CIRCLE_SHA1" 
        "$CIRCLE_TAG" 
        "$CIRCLE_PROJECT_USERNAME" 
        "$CIRCLE_PROJECT_REPONAME" 
        "$CIRCLE_BUILD_URL"
        > version.json
```

Dockerfile-build
```diff
@@ -0,0 +1,22 @@
+FROM node:0.10.45
```
mostlygeek: Is there a reason to name this file Dockerfile-build instead of just Dockerfile?

jbuck:
@seanmonstar @mostlygeek will take a look at it!

Dockerfile-build
jbuck: There's an existing Dockerfile that's used for fxa self-hosting - I think we'll look at tackling the self-hosting/local dev after we get it working in prod

PR #191, opened 2016-05-26
fix(config) Use OS temporary directory for local image uploads

jbuck:


vladikoff:
cc @seanmonstar

rfk:
@jbuck can you say more about the motivation for this change?  IIUC it means that the uploaded images won't survive e.g. a machine restart.  That doesn't matter for us in production, but would be bad for self-hosters who might be running this code.

jbuck:
I think I ran into some permissions issues with the Dockerfile that I managed to fix another way. If this breaks self-hosting then I agree that this shouldn't be merged

PR #197, opened 2016-07-21
fix(dev): adjust static resource script

vladikoff:
@seanmonstar r?


vladikoff:
won't help, different problem

PR #198, opened 2016-07-21
chore(config): Remove unused git key from config

jbuck:


jbuck:
@jrgm r?

jrgm:
Yep. I don't see it used anywhere.

PR #204, opened 2016-07-25
fix(config): only allow https gravatars

vladikoff:
Fixes https://github.com/mozilla/fxa-content-server/issues/3950


vladikoff:
@seanmonstar r?

seanmonstar:
This could easily just be in the prod config, right?

vladikoff:
> This could easily just be in the prod config, right?

True but why prod only?

seanmonstar:
Since that's where the current issue is. The default regex doesn't look like an accident (I would used a simpler one unless someone told me not to), and I'm hesitant to break the use case that existed for it.

vladikoff:
> Since that's where the current issue is. The default regex doesn't look like an accident (I would used a simpler one unless someone told me not to), and I'm hesitant to break the use case that existed for it.

Hm, `http` avatars will not work anywhere with the current setting due to CSP rules of the content server. @rfk thoughts?

shane-tomlinson:
> I'm hesitant to break the use case that existed for it.

Do you remember what the use case is/was?

I have to admit that mixed content is not something I think we should allow unless we have a very good reason to. Who knows what kind of junk could be served to our site by any goon in the middle. I don't know how it'd be exploitable, but do we want to allow it? Doesn't the browser block mixed content unless the user clicks to allow it?

vladikoff:
TODO: update API.md docs to switch to https urls

seanmonstar:
> Do you remember what the use case is/was?

Nope, no one seems to. Which is what worries me. As I said, the original regex doesn't look like an accident: it's complicated enough that there likely *was* a reason for it. I worry that if we just turn it off, we perhaps may start having issues, such as when we turned on flowId.

> I have to admit that mixed content is not something I think we should allow unless we have a very good reason to. Who knows what kind of junk could be served to our site by any goon in the middle. I don't know how it'd be exploitable, but do we want to allow it? Doesn't the browser block mixed content unless the user clicks to allow it?

Yep! Mixed content is the suck.

I wonder, how are these http urls being submitted anyways? Through the settings page, or something else that we don't control is POSTing to the profile server?

rfk:
This regex sprang into existence in https://github.com/mozilla/fxa-profile-server/commit/f9c054a5cac8ffcaf520a133f58a196faa2e151b, which sadly doesn't give any clues about the reason behind it.  I don't recall any of the particulars - if I had to guess, I'd guess that it was for testing purposes.

If we can confirm that we do not, in fact, have any http:// gravatar URLs in the production database, my suggest would be to remove support for them as in this patch, let whatever tests or dev environments that are going to fail fail, and we can see what the original problem may have been.

Of course, if we *do* have such URLs in the production database, we've got more work to do in figuring out why and what to do about it...

vladikoff:
Okay only secure images are used in the DB: https://github.com/mozilla/fxa-content-server/issues/3950#issuecomment-235331226 so this is good to go!

PR #215, opened 2016-10-29
chore(config): remove obsolete awsbox config file

jrgm:
I can't see any references to this file. Seems like an orphan that can be removed.


jrgm:
By the way, the other awsbox was removed in https://github.com/mozilla/fxa-profile-server/commit/d2d2bbecc692097a75daee7269684b3e035dbe32

PR #216, opened 2016-10-31
fix(config): log config at info level at startup

jrgm:
This is already shown in dev mode at log level debug. But this will show it when not at debug.


vladikoff:
👍

PR #220, opened 2016-11-10
fix(docker): Shrink docker image size

jbuck:
The docker image is currently **819.5 MB**. By switching to alpine linux we can reduce the size of the image to **122.8 MB**. By splitting the docker image into a production-only image we can shrink it down to **68.04 MB**.


jbuck:
Note to self: need to change the jenkins v2 run command

vladikoff:
from mtg: what is this blocked on?

jbuck:
@vladikoff nothing!

r? anyone

vladikoff:
alpine linux it is!

PR #221, opened 2016-11-14
feat(mysql): Ensure db connections always run in strict mode.

rfk:
This was an idea I had to enforce mysql strict mode and utf8mb4 on all db connections.  It's possible (and advisable!) to set these defaults in the MySQL server itself, but we can also enforce them here as an additional layer of protection.

No tests yet, because we don't have any test infra for mocking out MySQL in this repo.

@jrgm @vladikoff thoughts?

### DIFF
--- a/lib/db/mysql/index.js
+++ b/lib/db/mysql/index.js
@@ -13,12 +13,19 @@ const logger = require('../../logging')('db.mysql');
 const P = require('../../promise');
 const patch = require('./patch');
 
+const REQUIRED_SQL_MODES = [
+  'STRICT_ALL_TABLES',
+  'NO_ENGINE_SUBSTITUTION',
+];
+const REQUIRED_CHARSET = 'UTF8MB4_UNICODE_CI';
+
 
 function MysqlStore(options) {
-  if (options.charset && options.charset !== 'UTF8_UNICODE_CI') {
-    logger.warn('createDatabase', { charset: options.charset });
+  if (options.charset && options.charset !== REQUIRED_CHARSET) {
+    logger.error('createDatabase.invalidCharset', { charset: options.charset });
+    throw new Error('You cannot use any charset besides ' + REQUIRED_CHARSET);
   } else {
-    options.charset = 'UTF8_UNICODE_CI';
+    options.charset = REQUIRED_CHARSET;
   }
   options.typeCast = function(field, next) {
     if (field.type === 'TINY' && field.length === 1) {
@@ -202,11 +209,11 @@ MysqlStore.prototype = {
   },
 
   _write: function _write(sql, params) {
-    return this._query(this._pool, sql, params);
+    return this._query(sql, params);
   },
 
   _read: function _read(sql, params) {
-    return this._query(this._pool, sql, params);
+    return this._query(sql, params);
   },
 
   _readOne: function _readOne(sql, params) {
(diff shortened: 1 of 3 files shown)


lib/db/mysql/index.js
```diff
@@ -27,6 +33,30 @@ function MysqlStore(options) {
     return next();
   };
   this._pool = mysql.createPool(options);
+
+  // Enforce sane defaults on every new connection.
+  // These *should* be set by the database by default, but it's nice
+  // to have an additional layer of protection here.
+  this._pool.on('connection', function(connection) {
```
rfk: In theory the overhead of doing this won't be too bad, because we only do it when we create a new connection for the pool, and we then re-use that connection for many queries.

One downside of doing this by listening to an event though, is that we can't block until the query has been completed.  So there'd probably be a race condition here, where calling code could use the connection in between the `SELECT @@sql_mode` and the `SET SESSION sql_mode`.

rfk: It wasn't obvious to me whether all the various helper methods used `_getConnection` though, it looks like e.g. `_query()` just directly takes a connection from the pool.  Is `_getConnection` intended to replace the pool's default connection-getting logic uniformly?

seanmonstar: Oh interesting. It seems we could make use of `_getConnection` inside `_query`, and then yes, we can be sure the session is always set correctly.

Currently, it looks like `_getConnection` was written so that we can release the connection after using it, but that in most cases we just use `pool.query` which keeps the connection internal and thus we never worry about releasing it.

rfk: Also, ISTM that doing this logic in `_getConnection` would cause it to be run on every query, while using the pool event means it runs only when we create a new connection.

seanmonstar: We could set a flag, like `connection._fxa_strict = true`, to only do it once per connection.

lib/db/mysql/index.js
```diff
@@ -13,12 +13,18 @@ const logger = require('../../logging')('db.mysql');
...
+const REQUIRED_SQL_MODES = [
+  'STRICT_ALL_TABLES',
+  'NO_ENGINE_SUBSTITUTION',
+];
+
 
 function MysqlStore(options) {
-  if (options.charset && options.charset !== 'UTF8_UNICODE_CI') {
-    logger.warn('createDatabase', { charset: options.charset });
+  if (options.charset && options.charset !== 'UTF8MB4_UNICODE_CI') {
```
vladikoff: nit: extract `'UTF8MB4_UNICODE_CI'` into `const REQUIRED_CHARSET = ...`;

lib/db/mysql/index.js
vladikoff: 👍

vladikoff:
@jrgm any other feedback?

rfk:
I've updated this based on discussion, and added some simple tests.  @seanmonstar r?

lib/db/mysql/index.js
```diff
@@ -13,12 +13,19 @@ const logger = require('../../logging')('db.mysql');
 const P = require('../../promise');
 const patch = require('./patch');
 
+const REQUIRED_SQL_MODES = [
+  'STRICT_ALL_TABLES',
+  'NO_ENGINE_SUBSTITUTION',
+];
+const REQUIRED_CHARSET = 'UTF8MB4_UNICODE_CI';
```
seanmonstar: Yay moving to `const`.

rfk:
@jrgm per discussion in the meeting, the implications of strict mode are described here:

  http://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sql-mode-strict

I don't see anything to worry about in there, and TBH if we *are* doing something that's affected by strict mode, I consider it a bug we need to uncover and fix ASAP.  This works well in travis and my local testing, so I'm going to optimistically merge it and keep an eye on behaviour in latest.

PR #224, opened 2016-11-24
chore(shrinkwrap): add npm script for shrinkwrap

### LINKED ISSUE #223: Add an `npm run shrinkwrap` command
It should "do the right thing" so we can stop thinking about it.

Ref 
https://github.com/mozilla/fxa-content-server/issues/4439
https://github.com/mozilla/fxa-auth-server/issues/1564
https://github.com/mozilla/fxa-profile-server/issues/223
https://github.com/mozilla/fxa-auth-db-mysql/issues/189
https://github.com/mozilla/fxa-customs-server/issues/149

Example: https://github.com/mozilla/fxa-auth-mailer/blob/master/package.json#L10

TDA:
This commit adds an npm script for generating shrinkwrap.
Fixes #223 
This one is by far the smallest change in shrinkwrap :D

@vladikoff and @shane-tomlinson


TDA:
@vladikoff updated this too :)

PR #226, opened 2016-12-19
fix(avatars): only delete avatars if avatars set

vladikoff:
@jrgm r?


jrgm:
I vaguely r+, but really maybe @seanmonstar should r=.

PR #227, opened 2017-01-05
fix(config): make NODE_ENV consistent across servers

vladikoff:
@mozilla/fxa-devs r?

Related to https://github.com/mozilla/fxa/issues/57

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -39,8 +39,8 @@ const conf = convict({
     arg: 'node-env',
     doc: 'The current node.js environment',
     env: 'NODE_ENV',
-    format: ['dev', 'test', 'stage', 'prod'],
-    default: 'dev'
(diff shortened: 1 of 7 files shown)


lib/config.js
```diff
@@ -39,8 +39,8 @@ const conf = convict({
     arg: 'node-env',
     doc: 'The current node.js environment',
     env: 'NODE_ENV',
-    format: ['dev', 'test', 'stage', 'prod'],
-    default: 'dev'
+    format: ['development', 'test', 'stage', 'production'],
+    default: 'development'
```
vbudhram: Looks like auth-server and content-server default to `production`. Should this default to it as well?

vladikoff: Good plan!

PR #236, opened 2017-02-14
chore(docker): switch from npm shrinkwraps to yarn

seanmonstar:
There may be something else I'm missing, but passes travis and circle!

cc @jbuck


Dockerfile-build
```diff
@@ -1,4 +1,4 @@
-FROM mhart/alpine-node:4.7.2
+FROM mhart/alpine-node:base-4.7.2
```
vladikoff: what's this base- difference?

package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "outdated": "npm outdated --depth 0",
     "preinstall": "./scripts/check_gm.sh",
-    "shrinkwrap": "npmshrink",
+    "shrinkwrap": "yarn install",
```
vladikoff: how does that work?

Dockerfile-build
```diff
@@ -9,14 +9,15 @@ ADD https://s3.amazonaws.com/dumb-init-dist/v1.2.0/dumb-init_1.2.0_amd64 /usr/lo
 RUN chmod +x /usr/local/bin/dumb-init
 ENTRYPOINT ["/usr/local/bin/dumb-init", "--"]
 
+RUN apk add --repository http://dl-cdn.alpinelinux.org/alpine/edge/testing/ --no-cache yarn
```
vladikoff: 👍

Dockerfile-build
seanmonstar: The tag `4.7.2` means nodejs@4.7.2 plus npm. The `base-` prefix means only the node binary, without npm.

The full details can be found in https://hub.docker.com/r/mhart/alpine-node/

package.json
seanmonstar: Yarn will read the `package.json`, check if any of the dependencies have been updated, and install them if so. It finishes by updating the `yarn.lock`. More at https://yarnpkg.com/en/docs/cli/install

Actually, we wouldn't want to update the lock file in bulk any more, except for occasions where bulk updating dependencies is actually your goal. Instead, we would use `yarn upgrade package@version`, which will update the `package.json` and the `yarn.lock`, but only for that package (and it's dependencies).

Dockerfile-build
vladikoff: oh cool!

package.json
```diff
@@ -49,6 +49,7 @@
     "jscs-jsdoc": "2.0.0",
     "load-grunt-tasks": "3.5.0",
     "mkdirp": "0.5.1",
+    "mocha": "^3.2.0",
     "mocha-text-cov": "0.1.1",
     "nock": "8.0.0",
     "npmshrink": "1.0.1",
```
vladikoff: i guess this needs to be removed? are we keeping the shrinkwrap file in the repo? if we are it might be useful to still copy that file into the docker image

README.md
```diff
@@ -12,40 +12,41 @@ Such as name, avatar, location, age, gender, [etc](https://wiki.mozilla.org/Iden
...
 
 ```
-npm test
+yarn test
 ```
 
 To run tests via Docker:
 ```
-docker run --rm -v $PWD:/opt/fxa --net=host mozilla/fxa_profile_server npm test
+docker run --rm -v $PWD:/opt/fxa --net=host mozilla/fxa_profile_server yarn test
```
vladikoff: :o

vladikoff: ahh no npm in the docker container, woosh...

seanmonstar:
I've removed the shrinkwrap, updated the README to replace usages of npm with yarn, and added a preinstall script that tries to detect if not being run by yarn.

vladikoff:
* need to update fxa-dev
* need to update fxa-local-dev
* train-82?

vladikoff:
There is a Dockerfile that can be removed I Think, it still uses npm: 

image

package.json
```diff
@@ -5,8 +5,8 @@
   "description": "Firefox Accounts Profile service.",
   "scripts": {
     "outdated": "npm outdated --depth 0",
```
vladikoff: can this be updated to use `yarn outdated`?  I don't think you need the `--depth` part

.travis.yml
```diff
@@ -26,11 +26,9 @@ env:
   - NODE_ENV=test DB=mysql IMG=aws
 
 before_install:
-  - npm config set spin false
   - "mysql -e 'DROP DATABASE IF EXISTS fxa_profile;'"
   - "mysql -e 'CREATE DATABASE fxa_profile;'"
 
 script:
-  - npm run outdated
```
vladikoff: this can run `yarn outdated`

package.json
vladikoff: actually i think we can just drop this script and use https://github.com/mozilla/fxa-profile-server/pull/236/files#r102555395

package.json
```diff
@@ -49,9 +49,9 @@
     "jscs-jsdoc": "2.0.0",
     "load-grunt-tasks": "3.5.0",
     "mkdirp": "0.5.1",
+    "mocha": "^3.2.0",
```
vladikoff: should this be just "3.2.0" like the rest of dependencies?

vladikoff:
Closing for now, npm 5 also added a lockfile thing so that might work *shrugs*

PR #237, opened 2017-02-23
chore(docker): Update to node v4.8.0

jbuck:


vladikoff:
travis is using this already, should be good!

PR #241, opened 2017-03-17
feat(push): notify the auth-server when profile updated

eoger:
Connects to https://github.com/mozilla/fxa-features/issues/118

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -221,6 +221,20 @@ const conf = convict({
       env: 'WORKER_URL'
     }
   },
+  authServerMessaging: {
+    region: {
+      doc: 'The region where the queues live',
+      format: String,
+      env: 'AUTH_SERVER_MESSAGING_REGION',
+      default: ''
+    },
+    profileUpdatesQueueUrl: {
+      doc: 'The queue URL to use (should include https://sqs.<region>.amazonaws.com/<account-id>/<queue-name>)',
+      format: String,
+      env: 'PROFILE_UPDATES_QUEUE_URL',
+      default: ''
+    }
+  },
   hpkpConfig: {
     enabled: {
       default: false,
--- a/lib/routes/avatar/upload.js
+++ b/lib/routes/avatar/upload.js
@@ -10,6 +10,7 @@ const config = require('../../config');
 const db = require('../../db');
 const hex = require('buf').to.hex;
 const img = require('../../img');
+const notifyProfileUpdated = require('../../updates-queue');
 const validate = require('../../validate');
 const workers = require('../../img-workers');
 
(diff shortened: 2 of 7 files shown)


lib/routes/avatar/upload.js
```diff
@@ -55,6 +59,7 @@ module.exports = {
         return db.addAvatar(id, uid, url, FXA_PROVIDER);
       })
       .done(function uploadDone() {
+        profileUpdatedQueue.send({ uid: uid.toString('hex') });
```
seanmonstar: It looks like this returns a `Promise`, over in the `sqs.js` file. Should we wait on it before replying, or is ignoring it on purpose? If it is, may be worth a comment acknowledging that so Future Us can remember.

eoger: This is on purpose, will add a comment.

lib/config.js
```diff
@@ -221,6 +221,20 @@ const conf = convict({
...
+      env: 'AUTH_SERVER_MESSAGING_REGION',
+      default: ''
+    },
+    profileUpdatesQueueUrl: {
+      doc: 'The bounce queue URL to use (should include https://sqs.<region>.amazonaws.com/<account-id>/<queue-name>)',
+      format: String,
+      env: 'PROFILE_UPDATES_QUEUE_URL',
+      default: ''
+    }
+  },
```
rfk: Let's get @jbuck and @jrgm input on this as well; guys, thoughts on using an SQS queue here to relay notifications back to the auth-server, and from there on to the connected devices?

lib/routes/avatar/upload.js
```diff
@@ -7,16 +7,20 @@ const assert = require('assert');
...
+const SQSSender = require('../../sqs')(logger);
 const validate = require('../../validate');
 const workers = require('../../img-workers');
 
 const FXA_PROVIDER = 'fxa';
 const FXA_URL_TEMPLATE = config.get('img.url');
 assert(FXA_URL_TEMPLATE.indexOf('{id}') !== -1,
     'img.url must contain the string "{id}"');
+const profileUpdatedQueue = new SQSSender(config.get('authServerMessaging.region'),
+                                          config.get('authServerMessaging.profileUpdatesQueueUrl'));
```
rfk: I wonder if we could push this down a level of abstraction, so we can create this `SQSSender` object in one place, and then all these other modules just have to import something and pass it the target uid.

eoger:
Thanks Ryan I put the queue creation in another module.

lib/config.js
seanmonstar: I don't recall, did we get an answer? I want to say it was "yea, sure, do it", but fuzzy.

eoger: IIRC, @rfk told me they discussed this in a meeting and the idea was accepted.

rfk: Yea, sure, do it. 👍

PR #247, opened 2017-05-17
refactor(avatar): remove POST avatar

### LINKED ISSUE #244: Remove support for /avatar/post
Gravatar support was removed from the content server in https://github.com/mozilla/fxa-content-server/pull/4927, we no longer use this endpoint.

Should we remove the endpoint here too?

vladikoff:
Fixes #244

@seanmonstar // @shane-tomlinson r?


vladikoff:
ping @seanmonstar @shane-tomlinson

PR #248, opened 2017-05-17
feat(db): support emoji in display name

vladikoff:
@rfk thoughts?


vladikoff:
image

jrgm:
sorta lgtm, but hard to know what should be valid. Haven't done the required reading for this exam.

jrgm:
I suppose it might be useful to toss in some of the very most common emojis, in addition to the extreme.

lib/routes/display_name/post.js
```diff
@@ -22,6 +22,14 @@ const EMPTY = Object.create(null);
...
...4\u2615\u2618\u261D\u2620\u2622\u2623\u2626\u262A\u262E\u262F\u2638-\u263A\u2640\u2642\u2648-\u2653\u2660\u2663\u2665\u2666\u2668\u267B\u267F\u2692-\u2697\u2699\u269B\u269C\u26A0\u26A1\u26AA\u26AB\u26B0\u26B1\u26BD\u26BE\u26C4\u26C5\u26C8\u26CE\u26CF\u26D1\u26D3\u26D4\u26E9\u26EA\u26F0-\u26F5\u26F7-\u26FA\u26FD\u2702\u2705\u2708-\u270D\u270F\u2712\u2714\u2716\u271D\u2721\u2728\u2733\u2734\u2744\u2747\u274C\u274E\u2753-\u2755\u2757\u2763\u2764\u2795-\u2797\u27A1\u27B0\u27BF\u2934\u2935\u2B05-\u2B07\u2B1B\u2B1C\u2B50\u2B55\u3030\u303D\u3297\u3299]|\uD83C[\uDC04\uDCCF\uDD70\uDD71\uDD7E\uDD7F\uDD8E\uDD91-\uDD9A\uDDE6-\uDDFF\uDE01\uDE02\uDE1A\uDE2F\uDE32-\uDE3A\uDE50\uDE51\uDF00-\uDF21\uDF24-\uDF93\uDF96\uDF97\uDF99-\uDF9B\uDF9E-\uDFF0\uDFF3-\uDFF5\uDFF7-\uDFFF]|\uD83D[\uDC00-\uDCFD\uDCFF-\uDD3D\uDD49-\uDD4E\uDD50-\uDD67\uDD6F\uDD70\uDD73-\uDD7A\uDD87\uDD8A-\uDD8D\uDD90\uDD95\uDD96\uDDA4\uDDA5\uDDA8\uDDB1\uDDB2\uDDBC\uDDC2-\uDDC4\uDDD1-\uDDD3\uDDDC-\uDDDE\uDDE1\uDDE3\uDDE8\uDDEF\uDDF3\uDDFA-\uDE4F\uDE80-\uDEC5\uDECB-\uDED2\uDEE0-\uDEE5\uDEE9\uDEEB\uDEEC\uDEF0\uDEF3-\uDEF8]|\uD83E[\uDD10-\uDD3A\uDD3C-\uDD3E\uDD40-\uDD45\uDD47-\uDD4C\uDD50-\uDD6B\uDD80-\uDD97\uDDC0\uDDD0-\uDDE6])\uFE0F/g;
```
rfk: I wonder how terrible it would be if we just removed `\uD8000-\uDFFFF` from the exclusion regex above, allowing all surrogate pairs.  Are their surrogate-pair characters that we explicitly don't want to allow?  I worry that if we just allow emoji, we'll still be excluding e.g. a variety of asian script symbols.

rfk:
> For completeness we'll also need to:

Oh, and also:

* Update `./lib/db/mysql/index.js` to add `SET NAMES utf8mb4 COLLATE utf8mb4_bin;` or similar in the same way that we enforce strict-mode on a per-connection basis.

lib/routes/display_name/post.js
rfk: Although we may want to reject invalid surrogate sequences.

rfk: > Are their surrogate-pair characters that we explicitly don't want to allow?

Hrm, my read of [1] suggests that we may want to disallow planes 3 and above because they don't serve any display-related purpose.  Perhaps we can find a surrogate-pair regex that will allow only the BMP and planes 1 & 2?  Unicode is hard...

[1] https://en.wikipedia.org/wiki/Plane_(Unicode)

jbuck:
I think we should also add some more characters from Asian chracter sets. Maybe some from Unicode Plane 3? https://en.wikibooks.org/wiki/Unicode/Character_reference/2F000-2FFFF

We'll also need to modify https://github.com/mozilla/fxa-profile-server/blob/b10b88ce544ef9777c16a49b84c2aeb5da0cf322/lib/db/mysql/index.js#L20 to UTF8MB4_BIN .

vladikoff:
train 89 , cc @jbuck

lib/routes/display_name/post.js
```diff
@@ -15,13 +15,20 @@ const EMPTY = Object.create(null);
 //   \u007F         - ascii DEL character
 //   \u0080-\u009F  - C1 (ansi escape) control characters
 //   \u2028-\u2029  - unicode line/paragraph separator
-//   \uD800-\uDFFF  - non-BMP surrogate pairs
```
vladikoff: from Mathias: 

>     // \uD800-\uDFFF - non-BMP surrogate pairs
> …which means that no astral symbols are allowed. As long as that part remains, you’re not making use of `utf8mb4`. You could consider omitting that part, but then again you might still want to block lone surrogates…

vladikoff:
@rfk are we missing anything else from this PR? r?

lib/routes/display_name/post.js
rfk: If we're removing it from the comment, should it also be removed from the definition of `ALLOWED_DISPLAY_NAME_CHARS`?

lib/routes/display_name/post.js
```diff
@@ -30,7 +37,10 @@ module.exports = {
   },
   validate: {
     payload: {
-      displayName: Joi.string().required().allow('').regex(ALLOWED_DISPLAY_NAME_CHARS)
+      displayName: Joi.alternatives().try(
+        Joi.string().required().allow('').regex(ALLOWED_DISPLAY_NAME_CHARS),
+        Joi.string().required().allow('').regex(ALLOWED_EMOJI_CHARS)
+      )
```
rfk: I don't understand the use of `try` here; IIUC this will either allow the same display-name chars as we did before, or it will allow display names composed *entirely* of emoij, but not something like `"embedded 👍 emoji"`.  Am I missing something?

test/api.js
```diff
@@ -941,7 +941,12 @@ describe('/display_name', function() {
       var NAMES = [
         'André Citroën',
         'the unblinking ಠ_ಠ of ckarlof',
-        'abominable ☃'
+        'abominable ☃',
+        // emoji
+        '👍',
+        '👍🏼',
+        '蚋',
+        '鱑'
```
rfk: I think we should add some mixed ascii/emoji testcases as well.

GitCop:
Thanks for contributing! Unfortunately, I'm here to tell you there were the following style issues with your Pull Request:

 * Commit: 030fdc40f1a12d1fa91bb1c95329966477b07901
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**

Guidelines are available at https://github.com/mozilla/fxa-profile-server

---
This message was auto-generated by https://gitcop.com

lib/routes/display_name/post.js
```diff
@@ -15,13 +15,12 @@ const EMPTY = Object.create(null);
...
 //   \u0080-\u009F  - C1 (ansi escape) control characters
 //   \u2028-\u2029  - unicode line/paragraph separator
-//   \uD800-\uDFFF  - non-BMP surrogate pairs
 //   \uE000-\uF8FF  - BMP private use area
 //   \uFFF9-\uFFFF  - unicode "specials" block
 //
 // We might tweak this list in future.
 
-const ALLOWED_DISPLAY_NAME_CHARS = /^(?:[^\u0000-\u001F\u007F\u0080-\u009F\u2028-\u2029\uD800-\uDFFF\uE000-\uF8FF\uFFF9-\uFFFF])*$/;
+const ALLOWED_DISPLAY_NAME_CHARS = /^(?:[^\u0000-\u001F\u007F\u0080-\u009F\u2028-\u2029\uE000-\uF8FF\uFFF9-\uFFFF])*$/;
```
vladikoff: https://en.wikipedia.org/wiki/UTF-16#U.2BD800_to_U.2BDFFF

GitCop:
Thanks for contributing! Unfortunately, I'm here to tell you there were the following style issues with your Pull Request:

 * Commit: 030fdc40f1a12d1fa91bb1c95329966477b07901
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**

Guidelines are available at https://github.com/mozilla/fxa-profile-server

---
This message was auto-generated by https://gitcop.com

lib/routes/display_name/post.js
vladikoff: > hey will never be assigned a character, so there should be no reason to encode them. 

So it should be "safe"

GitCop:
Thanks for contributing! Unfortunately, I'm here to tell you there were the following style issues with your Pull Request:

 * Commit: 030fdc40f1a12d1fa91bb1c95329966477b07901
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**

Guidelines are available at https://github.com/mozilla/fxa-profile-server

---
This message was auto-generated by https://gitcop.com

vladikoff:
cc @jrgm ^

Issue #160, opened 2015-10-30
Allow only a single avatar per user


rfk:
We currently allow users to upload multiple avatars, with the original idea of allowing them to switch between pictures on e.g. a relier by relier basis.  We haven't shipped that and we have no concrete plans to, so we're paying a complexity cost that doesn't give us any benefit.

Let's simplify our backend avatar support so that the user only has a single avatar, and uploading a new one deletes the old one.

Also as noted in https://github.com/mozilla/fxa-profile-server/issues/94, this will make cleanup of old avatar pictures easier, and will probably help clients with discovering that the avatar has changed.

jstrutz:
@rfk I'm happy to take a crack at this, if it is still on the table?

rfk:
Sure!  I don't think it should be *too* complicated a cleanup; @seanmonstar are you happy to provide a bit of mentorship here?

seanmonstar:
Absolutely. Let me know if you need a breakdown, or if you feel comfortable just chopping, go right ahead.

jstrutz:
Great!  I'd be happy to hear your breakdown @seanmonstar .  How far do we want to take the cleanup? - i.e. do we want to simplify the API?  Change the DB schema?

seanmonstar:
- we can drop the [avatar_selected](https://github.com/mozilla/fxa-profile-server/blob/master/lib/db/mysql/schema.sql#L25) table
- remove the [selected](https://github.com/mozilla/fxa-profile-server/blob/master/lib/db/mysql/index.js#L154) argument to `db.addAvatar`
- remove the [getSelectedAvatar and getAvatars](https://github.com/mozilla/fxa-profile-server/blob/master/lib/db/mysql/index.js#L177-L183) db methods
- Mirror those chops with the memory db
- remove the [selected](https://github.com/mozilla/fxa-profile-server/blob/master/lib/routes/avatar/post.js#L32) parameter from the `POST` route (probably just make it optional for now and don't check it in the handler)
- The `routes/list.js` should just get the axe.

rfk:
> we can drop the avatar_selected table

Also to flag, this sounds like the sort of thing that's going to require at least two deployments to do in a rollback-compatible way, i.e. we can only actually drop it *after* there's no more code using it running in production.

vladikoff:
[...]

PR #267, opened 2017-07-07
feat(avatar): allow deleting the default avatar by not sending an id

seanmonstar:
Sending `DELETE /v1/avatar` will find the 'selected' avatar for the user
and delete that. A separate request to get the avatar ID is no longer
necessary.

This is related to #160, but I finally got around to doing it to reduce the number of requests the content-server needs to make. It has been fetch `GET /v1/profile/avatar` specifically so that it can have an ID ready in case the user clicks to delete their avatar. Save a request!


vladikoff:
Also related to https://github.com/mozilla/fxa-profile-server/issues/160

Issue #261, opened 2017-06-28
display name length error


udaraweerasinghege:
If you create a display name that is very long (not sure what allowed length is), it will save successfully, however if you refresh the page, the following error can be seen in profile server:

[...]

PR #275, opened 2017-07-26
fix(displayName): length validation in post

udaraweerasinghege:
Fixes https://github.com/mozilla/fxa-profile-server/issues/261
cc: @vladikoff


vladikoff:
@udaraweerasinghege thanks so much!!

PR #282, opened 2017-09-14
feat(cache): Delete user cache on email change

### LINKED ISSUE #280: Changing primary email does not invalidate cache
Need to find a way to signal from the auth server to the profile server redis that the email changed.

vbudhram:
This PR invalidates a user's cache when server receives a `primaryEmailChanged` message. To keep things simple, I opted to pass the server object into events and invalidate the cache from there.

Fixes #280 

@vladikoff @rfk does this approach seem reasonable? r?


lib/events.js
```diff
@@ -42,27 +42,74 @@ exports.onData = function onData(message) {
...
-    if (env.isProdLike()) {
-      throw new Error('config.events must be included in prod');
+    // We currently a generate a user's cache key from their
+    // credentials in a request object. This builds an object
+    // to match what the server is expecting.
+    var req = {auth: {credentials: {user: userId}}};
+    P.resolve(new Promise(function(resolve) {
+      server.methods.batch.cache.drop(req, function () {
+        logger.info('primaryEmailChanged:cacheCleared', {uid: userId});
+        notifyProfileUpdated(userId);
```
rfk: Calling `notifyProfileUpdated` here won't cause any harm, but  don't think it's necessary - its only job is to tell the auth-server when profile data changes so that it can end push notifications to connected devices, but it in this case it already knows about the change, so we'll end up sending the push notifications twice.  Adding it here will cause us to send that push notification twice (once when auth-server processes the email change, and again when it receives this notification from profile-server)

lib/events.js
```diff
@@ -42,27 +42,74 @@ exports.onData = function onData(message) {
...
 
-exports.start = function start() {
-  if (! config.events.region || ! config.events.queueUrl) {
-    if (env.isProdLike()) {
-      throw new Error('config.events must be included in prod');
+    // We currently a generate a user's cache key from their
+    // credentials in a request object. This builds an object
+    // to match what the server is expecting.
+    var req = {auth: {credentials: {user: userId}}};
+    P.resolve(new Promise(function(resolve) {
```
rfk: Can anything inside this `new Promise` block fail or throw, e.g. if redis is down or if there's an error talking to SQS?  It may be worth wrapping in a try-catch so that the returned promise will properly error out in such cases.

Also it's not clear why we need to `P.resolve(new Promise())`, we should just be able to use the object created by  `new Promise` directly?

And if it is necessary, we should use either `Promise` or `P` consistently rather than mixing the two :-)

lib/events.js
```diff
@@ -42,27 +42,74 @@ exports.onData = function onData(message) {
...
+      server.methods.batch.cache.drop(req, function () {
+        logger.info('primaryEmailChanged:cacheCleared', {uid: userId});
+        notifyProfileUpdated(userId);
+        resolve();
+      });
+    })).done(
```
rfk: I wonder if factoring out this `done(...message.del()...)` logic into some common routine would make the code clearer and more robust, e.g. by having `deleteUser` and `primaryEmailChanged` return a promise, and having `onData` add a common cleanup block onto it.

vbudhram:
@rfk Refactored code to make it a little cleaner and added an extra test for checking uids. Mind another r?

vbudhram:
@rfk Thank you!

PR #286, opened 2017-10-03
chore(ci): only test node 4, update nsp

vladikoff:


.nsprc
```diff
@@ -0,0 +1,6 @@
+{
+  "exceptions": [
+    // gm
+    "https://nodesecurity.io/advisories/534"
```
vladikoff: https://github.com/mozilla/fxa-profile-server/issues/287

PR #293, opened 2017-11-13
fix(logging): Don't log raw numbers as log msg.

### LINKED ISSUE #292: compute.image-cc.time.ms.gm.small event isn't compatible with mozlog
```
{"EnvVersion":"2.0","Severity":6,"Timestamp":"2017-11-03 14:09:12","Pid":5,"Fields":{"msg":"spawned new worker process (11) 1/5 processes running"},"Logger":"fxa-profile-server","Type":"compute.image-cc"}
{"EnvVersion":"2.0","Severity":6,"Timestamp":"2017-11-03 14:09:19","Pid":11,"Fields":{"msg":1649},"Logger":"fxa-profile-server","Type":"compute.image-cc.time.ms.gm.small"}
```

`"msg":1649` isn't compatible mozlog, it needs to be a string. We should also make sure the other resize events don't print a number

rfk:
Fixes #292; @jbuck r?

### DIFF
--- a/lib/compute/image-cc.js
+++ b/lib/compute/image-cc.js
@@ -64,11 +64,11 @@ function compute(msg, callback) {
   logger.debug('process.start', { bytes: src.length, variant: variant });
   processImage(src, msg.width, msg.height).then(function(out) {
     s3Start = Date.now();
-    logger.info('time.ms.gm.' + variant, s3Start - start);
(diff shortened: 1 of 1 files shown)


shane-tomlinson:
@rfk - is this for train-100?

rfk:
> is this for train-100?

It's tiny enough that I feel like it could be, let's touch on it in the waffleboard triage today.

PR #298, opened 2018-01-11
fix(config): mark config sentryDsn and mysql password sensitive

### LINKED ISSUE #297: bin/server prints mysql cleartext password on startup
When starting the server, for instance by using `NODE_ENV=development node bin/server.js`, an INFO message is printed containing the entire `config`, including the cleartext password for the mysql database, both in development and production environments. As this could easily end up in log files, it would probably be best to remove this confidential information from the output.

jrgm:
r? - @vladikoff

Updates node-convict and marks mysql password and sentryDsn as `sensitive`.

Output is now: `{... "mysql":{"createSchema":true,"user":"root","password":"[Sensitive]", ... "sentryDsn":"[Sensitive]"}`

Fixes #297


lib/config.js
```diff
@@ -89,6 +89,7 @@ const conf = convict({
       },
       types: {
         doc: 'A mapping of allowed mime types and their file signatures',
+        format: Object,
```
jrgm: convict@4.0.2 validation now requires this type information.

lib/config.js
```diff
@@ -346,7 +349,7 @@ if (conf.get('env') === 'test') {
 process.env.NODE_ENV = conf.get('env');
 
 var options = {
-  strict: true
+  allowed: 'strict'
```
jrgm: new syntax for the same thing

bin/server.js
```diff
@@ -6,13 +6,14 @@
...
-const config = require('../lib/config').getProperties();
+const configuration = require('../lib/config');
 const db = require('../lib/db');
 const logger = require('../lib/logging')('bin.server');
 const server = require('../lib/server').create();
 const events = require('../lib/events')(server);
 
-logger.info('config', config);
+// The stringify/parse is to force the output back to unindented json.
+logger.info('config', JSON.stringify(JSON.parse(configuration.toString())));
```
jrgm: Bit of a pain that `toString()` hardwires a `null, 2` stringify, but that's what it does.

Issue #295, opened 2017-12-20
Provide a default avatar URL if the user hasn't uploaded one?


rfk:
From discussion in https://github.com/mozilla-lockbox/lockbox-extension/issues/417

We have a nice "placeholder" avatar image that we use in our UI when the user hasn't uploaded a custom image.  For consistency it would be nice for reliers to be able to use the same image.

For users who have not set an avatar image, should we just return the URL of the placeholder image as part of the user's profile data?

pdehaan:
Q: Should it return the URL to the default avatar image, or inline as a data-uri in the response?

**Pro:** One less network request.
**Con:** Inlined response is much larger.

IvanDelsinne:
Returning an URL would require the client to make another request to fetch said avatar, but the client could also choose to not request the image. In the opposite case, in which the client would grab the user's avatar, if not the default, he'd still require it, so I'd say it'd be expected behaviour.

Having links in both situations would keep one concise response type: HTTP(s) URL to an image resource.

Also, if any client would want to not load the icon but provide a locally-stored version, they still could just check the URL against the def. avatar's link (or a list of default avatar's links).

For me, it'd be a yes.

vladikoff:
from mtg: let's do it! Let's make it a URL to the placeholder

PR #300, opened 2018-02-09

vladikoff:
Fixes https://github.com/mozilla/fxa-profile-server/issues/295

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -158,6 +158,11 @@ their equivalent) can be accessed as multiple sizes. The default size is
 variants, which can accessed by adding the `_small` or `_large` suffix
 to the avatar URL.
 
+If there is no avatar set by the user, then the server will provide a url to a default avatar image.
+Keep in mind that this default url will have an extension `.png`, unlike avatars that are set by users.
+You can access `_small` and `_large` versions by changing the file name to `default_small.png` and `default_large.png`
+
+
 #### Request
 
 ```sh
@@ -171,7 +176,17 @@ curl -v \
 ```json
 {
   "id": "81625c14128d46c2b600e74a017fa4a8",
-  "url": "https://secure.gravatar.com/avatar/6d940dd41e636cc156074109b8092f96"
+  "avatar": "https://secure.gravatar.com/avatar/6d940dd41e636cc156074109b8092f96"
+}
+```
+
+#### Response (no avatar set)
+
+```json
+{
+  "default": true,
+  "id": "00000000000000000000000000000000",
+  "avatar": "https://accounts.firefox.com/images/avatars/default.png"
 }
 ```
 
--- a/lib/config.js
+++ b/lib/config.js
@@ -133,6 +133,11 @@ const conf = convict({
       doc: 'Pattern to generate FxA avatar URLs. {id} will be replaced.',
       default: 'http://127.0.0.1:1112/a/{id}',
       env: 'IMG_URL'
+    },
+    defaultAvatar: {
+      doc: 'Location of the default avatar image',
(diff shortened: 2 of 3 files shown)


docs/API.md
```diff
@@ -158,6 +158,11 @@ their equivalent) can be accessed as multiple sizes. The default size is
 variants, which can accessed by adding the `_small` or `_large` suffix
 to the avatar URL.
 
+If there is no avatar set by the user, then the server will provide a url to a default avatar image.
+Keep in mind that this default url will have an extension `.png`, unlike avatars that are set by users.
+You can access `_small` and `_large` versions by changing the file name to `default_small.png` and `default_large.png`
```
vladikoff: @rfk thoughts on this madness?

docs/API.md
```diff
@@ -171,7 +176,17 @@ curl -v \
...
+}
+```
+
+#### Response (no avatar set)
+
+```json
+{
+  "default": true,
+  "id": "00000000000000000000000000000000",
+  "avatar": "https://accounts.firefox.com/images/avatars/default.png"
```
vladikoff: firefoxusercontent here? or should we still use the content server?

docs/API.md
```diff
@@ -171,7 +176,17 @@ curl -v \
...
-  "url": "https://secure.gravatar.com/avatar/6d940dd41e636cc156074109b8092f96"
+  "avatar": "https://secure.gravatar.com/avatar/6d940dd41e636cc156074109b8092f96"
+}
+```
+
+#### Response (no avatar set)
+
+```json
+{
+  "default": true,
```
vladikoff: would this be useful?

vladikoff:
from mtg: 

- we can probably change the api
- either stop using small and large OR provide those in the api response as `avatar_large` / `avatar_small`
- uploading default options: content server OR profile server uploads to firefoxusercontent OR manual upload

docs/API.md
rfk: Is it necessary for us to serve it with a `.png` extension, or would we just serve it at some well-known id?  Since we're doing this to simplify the work of reliers, sending something like https://firefoxusercontent.com/00000000000000000000000000000000 might be more useful than sending something they have to special-case.  (This would apply doubly-so to constructing the `_small` and `_large` variants).

docs/API.md
```diff
@@ -171,7 +176,17 @@ curl -v \
 ```json
 {
   "id": "81625c14128d46c2b600e74a017fa4a8",
-  "url": "https://secure.gravatar.com/avatar/6d940dd41e636cc156074109b8092f96"
+  "avatar": "https://secure.gravatar.com/avatar/6d940dd41e636cc156074109b8092f96"
```
rfk: Since we don't support gravatar, let's change the example text to a plausible URL.  Also, was the doc here in error with "url" vs "avatar"?

docs/API.md
rfk: On reflection, I think it's fine to just be `default` here because the URL is specifically asking for information about the avatar.  I guess this key would simply not show up in the response for the combined `/profile` endpoint?

rfk:
> either stop using small and large OR provide those in the api response as avatar_large / avatar_small

@jrgm do we have visibility into the request logs for firefoxusercontent.com.  If so, can we please investigate whether any reliers are requesting the "XXXXX_small" and "XXXXX_large" variants of our profile images?

rfk:
I'm 👍 to the approach here, we just have to figure out precisely how we want to host the default image.  Uploading it out-of-band may be the simplest approach, at the cost of having to remember to maintain that copy if we change the original in content-server.

PR #304, opened 2018-02-23

### LINKED ISSUE #295: Provide a default avatar URL if the user hasn't uploaded one?

vladikoff:
Fixes #295


vladikoff:
@rfk r?

Added information into the deploy doc after talking this through with @jrgm

docs/API.md
```diff
@@ -166,12 +166,36 @@ curl -v \
 "https://profile.accounts.firefox.com/v1/avatar"
 ```
 
-#### Response
+
+#### Response 
+
+> Note: we no longer support Gravatar uploading, but there are entries in the database
```
rfk: Are there?  Side note, but I wonder if we should somehow purge them.

lib/server/_static.js
```diff
@@ -20,6 +28,26 @@ exports.create = function() {
     port: config.server.port + 1
   });
 
+  server.route({
+    method: 'GET',
+    path: '/a/' +  DEFAULT_AVATAR_ID + '{type?}',
```
rfk: 👍  I like this way of just making it all a bit simpler for dev

docs/API.md
vladikoff: Mistake, we actually used to process them, I guess the docs always had an incorrect url

PR #306, opened 2018-02-28

vladikoff:


vladikoff:
* [ ] Need to add a default image flag to the api so content server knows when the image is set or not.

Issue #309, opened 2018-03-13
Clean up `Buffer` calls to deprecated API


vladikoff:
![image](https://user-images.githubusercontent.com/128755/37316284-85d39ae0-2634-11e8-8e40-67a9f67b4208.png)

cc @deeptibaghel

vladikoff:
Ref https://github.com/mozilla/fxa/issues/274

deeptibaghel:
@vladikoff i will complete it.

PR #310, opened 2018-03-13
fix(buffer): #309 Clean up 'Buffer' calls to deprecated API

deeptibaghel:
Fixes for #309 Clean up 'Buffer' calls to deprecated API


vladikoff:
@deeptibaghel Thank you!

(`Fixes for #309` won't be picked up by Github, remove the `for` and then it should work!)

PR #311, opened 2018-03-27
Report AMR info in profile data.

### LINKED ISSUE #302: Report account MFA status as part of profile info
For the IAM integration, they want to be able to query whether an account is MFA enabled. Github handles this by reporting a simple "two_factor_authentication": true|false field as part of the user's profile info, and we could do something similar here.

Depends on the auth-server exposing this data: mozilla/fxa-auth-server#2314

rfk:
Fixes #302.  I want to take a fresh look at this tomorrow, but putting it up for CI in the meantime.


docs/API.md
```diff
@@ -84,6 +84,9 @@ curl -v \
 {
   "uid": "6d940dd41e636cc156074109b8092f96",
   "email": "user@example.domain",
+  "locale": "en-US",
+  "amr_values": ["pwd", "otp"],
```
rfk: I'm torn here, between using "amr_values" for consistency with OAuth specs, versus "amrValues" for consistency with camel-casing of "avatarDefault" and "displayName".  I think I'll probably wind up favouring local consistency and switching these to camel-case.

lib/routes/profile.js
```diff
@@ -71,13 +69,6 @@ module.exports = {
           result.sub = creds.user;
         }
 
-        if (result.avatar) {
-          // currently the batch requests extract a single property.
-          // to avoid refactoring the batch requests to support multiple properties,
-          // set the default flag here
-          result.avatarDefault = result.avatar === DEFAULT_AVATAR_URL;
-        }
```
rfk: @vladikoff I wound up doing that refactor you were after :grin:

rfk:
OK, I think this is ready for an r?  There's a surprising amount of churn here, so let me break it down a bit it:

* The first commit lets us start using ES6 syntax, by migrating from blanket to nyc for code coverage.  Unfortunately this required a bit of mucking around with the database logic in order to get the tests passing.
* The second commit introduces the new profile info fields, which it needs to fetch from the auth-server.
    * The logic for fetching profile info from auth-server has been moved into an internal-only route named `/_core_profile`, since it now has several potential callers.
    * The existing `/email` route is now a thing wrapper over `/_core_profile`.
    * The `/profile` route now includes all fields from the auth-server via `/_core_profile`, rather than just the `email` field.

@vbudhram r?

docs/API.md
```diff
@@ -84,6 +84,9 @@ curl -v \
 {
   "uid": "6d940dd41e636cc156074109b8092f96",
   "email": "user@example.domain",
+  "locale": "en-US",
```
rfk: We get `locale` here for free as part of the broader refactor.

lib/batch.js
```diff
@@ -14,57 +14,63 @@ function inject(server, options) {
...
+// while mapping it to an array will include only the fields named
+// in the array.  For example:
 //
-// Ex: batch(req, { email: '/v1/email', avatar: '/v1/avatar' }).done(reply)
+//   batch(req, {
+//     // Include all fields returned by /v1/email
+//     '/v1/email': true
+//     // Include only two of the fields returned by /v1/avatar
+//     '/v1/avatar': ['avatar', 'avatarDefault']
+//   }).done(reply)
```
rfk: This is a significant change to how `batch` works; previously it could only fetch a single field from each route that it calls, and I've changed it so that it basically flattens all the fields into a single object.

lib/batch.js
```diff
@@ -14,57 +14,63 @@ function inject(server, options) {
...
+            email = res.result.email;
+          }
+          result[field] = res.result[field];
+        });
+        break;
       case 204:
       case 403:
       case 404:
-        logger.debug(prop + '.' + res.statusCode, {
+        logger.debug(url + ':' + res.statusCode, {
```
rfk: This is purely stylistic, but `/v1/avatar:401` feels better than `/v1/avatar.401` now that the keys are URL paths rather than just string properties like "avatar".

lib/db/index.js
```diff
@@ -10,68 +10,116 @@ const logger = require('../logging')('db');
...
+
+/**
+ * Clear database state and shutdown the active instance.
+ * This is used by tests during their teardown phase.
+ */
+exports._teardown = function _teardown() {
+  return exports._clear().then(() => {
+    return exports.finalize();
+  });
+};
```
rfk: As a side-effect of moving to nyc, the test runner would hang after completing all the tests when run with `DB=mysql`.  It turns out it was blocking on open connections in the MySQL connection pool, so I've added some explicit cleanup logic in the tests.  I also added some explanatory comments in this file because it look me a while to remember what it was doing.

lib/routes/_core_profile.js
```diff
@@ -0,0 +1,91 @@
...
+  },
+  response: {
+    schema: {
+      email: Joi.string().optional(),
+      locale: Joi.string().optional(),
+      amrValues: Joi.array().items(Joi.string().required()).optional(),
+      twoFactorAuthentication: Joi.boolean().optional()
+    }
+  },
+  handler: function _core_profile(req, reply) {
```
rfk: The logic here is coped pretty directly from the old `/email` route, including error-handling etc.

lib/routes/root.js
```diff
@@ -41,12 +41,13 @@ module.exports = {
...
+    const gitDir = path.resolve(__dirname, '..', '..', '.git');
+    exec('git rev-parse HEAD', { cwd: gitDir }, (err, stdout) => { // eslint-disable-line handle-callback-err
       commitHash = stdout.replace(/\s+/, '');
-      var configPath = path.join(gitDir, 'config');
-      var cmd = 'git config --get remote.origin.url';
-      exec(cmd, { env: { GIT_CONFIG: configPath } }, function(err, stdout) { // eslint-disable-line handle-callback-err
+      const configPath = path.join(gitDir, 'config');
+      const cmd = 'git config --get remote.origin.url';
+      const env = Object.assign({}, process.env, { GIT_CONFIG: configPath });
+      exec(cmd, { env }, (err, stdout) => { // eslint-disable-line handle-callback-err
```
rfk: For unknown reasons, the move to nyc meant that this started failing on my local machine.  I found I had to explicitly pass in `process.env` in order for it to locate the `git` executable correctly.

package.json
```diff
@@ -8,7 +8,7 @@
     "preinstall": "./scripts/check_gm.sh",
     "shrink": "npmshrink",
     "start": "scripts/run_dev.js",
-    "test": "grunt test --node-env=test"
+    "test": "scripts/test-local.sh"
```
rfk: This script, and the `mocha-coverage.js` script that it calls, were copied straight from oauth-server.

test/api.js
```diff
@@ -84,31 +84,31 @@ describe('/profile', function() {
...
           && rec.args[0] === 'summary'
-          && rec.args[1].path === '/v1/email';
+          && rec.args[1].path === '/v1/_core_profile';
     });
 
-    mock.log('batch', function(rec) {
+    mock.log('batch', rec => {
       return rec.levelname === 'ERROR'
-          && rec.args[0] === 'email.503';
+          && rec.args[0] === '/v1/_core_profile:503';
```
rfk: This was the result of the stylistic change to log output from earlier.

test/profileCache.js
```diff
@@ -64,12 +64,13 @@ function makeProfileReq(uid) {
...
 describe('profile cache', function() {
   beforeEach(() => {
-    Server = require('./lib/server');
     clearRequireCache();
+    Server = require('./lib/server');
   });
 
   afterEach(() => {
     mock.done();
+    require('../lib/db')._teardown();
```
rfk: This was a head-scratcher - the new test runner would hang when executing these tests, but only if they ran *after* some other tests.  I eventually realized doing `require('./lib/server')` and then `clearRequireCache`, was causing it to import a copy of `./lib/db` and then discard the reference to it.  We thus couldn't close its connection pool and would cause the tests to hang.  It's fixed by clearing the require cache before loading `Server`, and by tearing down the db module after each test rather than only once at the end.

I'm not thrilled about the way a bunch of implied shared state leaks between tests like this, but fixing that seemed like a bigger refactor than I was willing to take on...

lib/routes/_core_profile.js
```diff
@@ -0,0 +1,91 @@
...
+        result.email = body.email;
+      }
+      if (typeof body.locale !== 'undefined') {
+        result.locale = body.locale;
+      }
+      // Translate from internal terminology into OAuth-style terminology.
+      if (typeof body.authenticationMethods !== 'undefined') {
+        result.amrValues = body.authenticationMethods;
+      }
+      if (typeof body.authenticatorAssuranceLevel !== 'undefined') {
```
vbudhram: If no `authenticatorAssuranceLevel` set, any benefit to setting `result.twoFactorAuthentication` to false?

rfk: I thought about it a little, but I kinda like the symmetry of not reporting anything if the auth-server doesn't report anything.  This way if we happen to disable the AAL stuff over in auth-server due to e.g. a bug, the profile-server will end up saying "I don't know" rather than "no 2fa here".

PR #312, opened 2018-03-29
fix(node): Use Node.js v6.14.0

jbuck:


vbudhram:
The `should upload a new avatar` consistently fail locally for me, if someone else wants to double check.

vladikoff:
@vbudhram huh but not on CI?

PR #315, opened 2018-04-01
fix(mozlog): deprecation

deeptibaghel:


vladikoff:
Thank you @deeptibaghel !

PR #328, opened 2018-07-04
run npm ls --production with tradeoffs

jrgm:
Filing this PR for comments on tradeoffs, but not necessarily for merging. Happy to close it as a bad tradeoff.

The first commit adds some system dependencies needed for newrelic native dependencies to build, and tries to not bloat too much, but kinda fails by growing from v1.115.0's 161MB to 349MB. 188MB of bloat is trivial on disk, but does matter in network time when spinning up autoscaling instances. I could also remove the newrelic dependency, but I'd like to move forward to make use of it in production.

The second commit can run `npm ls --production`, but it requires the bloat. (If the newrelic deps fail, `npm ls --production` fails).

The third commit changes circle to use a machine executor and enabling `docker` experimental features so I can try out a `--squash`. That would actually reduce this PR's docker to be smaller than the current profile-server image. But it would require shipping production code on experimental docker builds, and may have some delay on build iteration time, and also possibly on security considerations.

So those are the tradeoffs. Could eat the bloat and get the `npm ls` checking. Or move to machine <s>experimental</s> executor with docker 17. 

Or wait for circleci docker builds to mature to more current versions.

Build sizes:
```
REPOSITORY:TAG                                                              SIZE
mozilla/fxa-profile-server:dockerpush-fix-run-npm-ls-production-and-squash  134MB
mozilla/fxa-profile-server:dockerpush-fix-run-npm-ls-production             349MB
mozilla/fxa-profile-server:v1.115.0                                         161MB
node:8-alpine                                                                68MB
```

r? - @mozilla/fxa-devs


jrgm:
> on experimental docker builds

Actually, not that experimental, since circleci/classic:edge has 17.10.0-ce and docker's stable release is 18.05.0-ce

jrgm:
Closing this in favour of #330.

PR #329, opened 2018-07-04
feat(scopes): Use shared code lib for checking OAuth scopes.

rfk:
Connects to mozilla/fxa-oauth-server#575

Replaces ad-hoc scope-checking logic with the shared code from mozilla/fxa-shared#21. WIP until that lands, as I want to iterate on it in conjunction with the consuming code.


lib/profileCache.js
```diff
@@ -11,41 +13,40 @@ const batch = require('./batch');
...
+  SCOPE_UNTRUSTED.toJSON()
+];
 
-// Extract the profile-related scopes for a request.
-// This removes non-profile scopes and returns either
-// the string 'profile' or a sorted plus-separated list
-// of subscopes like 'display_name+email'.
 
-function extractProfileScopes(req) {
+function getProfileCacheKey(req) {
```
rfk: The diff here got pretty awful, I I feel like the end result is not too bad.  It's slighly less "neat" than the previous version, but will do the same job and avoids having to parse a bunch of scopes here in this code.

lib/profileCache.js
```diff
@@ -11,41 +13,41 @@ const batch = require('./batch');
...
-    } else if (scope.startsWith('profile:')) {
-      subScopes[scope.substr('profile:'.length)] = true;
-    } else if (scope === 'email') {
-      subScopes.email = true;
+    // If they're using a set of scopes common to untrusted reliers,
+    // cache under a key for that specific set of scopes.  It's important
+    // that they don't have other profile-related scopes, otherwise we
+    // give them cached data that's missing some fields they're entitled to.
+    else if (scope.contains(SCOPE_UNTRUSTED) && ! scope.difference(SCOPE_UNTRUSTED).intersects(SCOPE_PROFILE)) {
+      creds._fxaProfileCacheKey = `${uid}:scoped:display_name+email+uid`;
```
shane-tomlinson: does `scoped` have special significance? What happens if someone requests e.g., `profile:display_name profile:email`, will that be cached under the same key? I'm having a hard time understanding fully 
`scope.difference(SCOPE_UNTRUSTED).intersects(SCOPE_PROFILE)`

lib/profileCache.js
```diff
@@ -11,41 +13,41 @@ const batch = require('./batch');
...
+      creds._fxaProfileCacheKey = uid;
     }
-    if (scope === 'profile') {
-      creds._fxaCachedProfileScopes = 'profile';
-      return 'profile';
-    } else if (scope.startsWith('profile:')) {
-      subScopes[scope.substr('profile:'.length)] = true;
-    } else if (scope === 'email') {
-      subScopes.email = true;
+    // If they're using a set of scopes common to untrusted reliers,
```
shane-tomlinson: Does this mean there has to be an exact match? If more profile scopes than are in the cache are requested, then the `! scope.difference(SCOPE_UNTRUSTED).intersects(SCOPE_PROFILE)` kicks in and says "no"?

lib/profileCache.js
rfk: > does scoped have special significance? 
No special significance, it's just a string to help clarify what's going on in the cache key name.
>  What happens if someone requests e.g., profile:display_name profile:email, will that be cached under the same key?

No, because the cached item contains the uid, which they don't have permission to view.

I'll see if I can add some more explanation in comments.

PR #330, opened 2018-07-09
Dockerpush fix docker use base builder run npm ls production

jrgm:
Okay,  instead of #328, this PR using a builder pattern, copying over /app from a base build that allows binary extensions to build.

The final production image is just 164MB or 3MB larger than mozilla/fxa-profile-server:v1.115.0 (161MB). (If I add a `--squash` to the build, the final images shrinks to 156MB, or 5MB below current production. But we can defer any discussion of whether to `--squash`, since it's not a big difference now).

r? - @vladikoff -- I'm going to close #328 in favour of this approach.


Dockerfile-test
```diff
@@ -1,2 +1,5 @@
 FROM fxa-profile-server:build
+USER root
+RUN apk add --repository http://dl-cdn.alpinelinux.org/alpine/v3.5/community/ --no-cache --virtual .build-deps git python make g++
```
jrgm: It isn't absolutely necessary to add the binary deps here, but if test dependencies have binary node modules, this would allow them to build.

Dockerfile-build
```diff
@@ -22,3 +23,24 @@ COPY scripts/check_gm.sh scripts/check_gm.sh
...
+# S3 bucket in Cloud Services prod IAM
+ADD https://s3.amazonaws.com/dumb-init-dist/v1.2.0/dumb-init_1.2.0_amd64 /usr/local/bin/dumb-init
+RUN chmod +x /usr/local/bin/dumb-init
+ENTRYPOINT ["/usr/local/bin/dumb-init", "--"]
+
+RUN apk add --repository http://dl-cdn.alpinelinux.org/alpine/v3.5/community/ --no-cache graphicsmagick git
+
+USER app
+
+COPY --from=builder --chown=app /app/ /app/
```
jrgm: If I don't say `--chown=app`, /app is copied over as `root:root`, and Dockerfile-test cannot do `npm install`.

jrgm: You might think that since `USER app`, the files would be copied as `app:app`, but it's not how `COPY --from` does it.

Dockerfile-test
vladikoff: yeap 👍

PR #334, opened 2018-07-31
[WIP] refactor(hapi): Upgrade to hapi 17

deeptibaghel:
Fixes #333
@vladikoff somehow npm start is not working, lib/server/_static.js seems to be having some issue. Could you pl. have a look.

PR #336, opened 2018-08-15
fix(test): Avoid cross test pollution caused by mutating route objects.

shane-tomlinson:
This is to help @deeptibaghel with #334. I added a ton of logging to Deepti's PR and Hapi and noticed that a route's `scope` value was being updated for every test suite, the next test suite would then use the implicant values from the previous suite as it's scopes.

This makes a copy of the route definition which is safe to mutate. Still a WIP because I would expect tests to be failing on master too, and I'm not sure why they aren't.

cc @rfk, @vladikoff, @deeptibaghel

### DIFF
--- a/lib/server/web.js
+++ b/lib/server/web.js
@@ -5,6 +5,7 @@
 const Hapi = require('hapi');
 const Raven = require('raven');
 const ScopeSet = require('fxa-shared').oauth.scopes;
+const cloneDeep = require('lodash.clonedeep');
 
 const AppError = require('../error');
 const config = require('../config').getProperties();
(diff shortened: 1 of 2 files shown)


vladikoff:
We need this to progress! 👍 🍰

vladikoff:
@deeptibaghel we need to find why the reason why this patch is needed in hapi 17 and not in the current version that we are using

shane-tomlinson:
I'd be happy for @deeptibaghel to use this to progress in other areas of her PR, but it makes me nervous that I still do not understand the root cause of why this is needed with hapi 17 and not the previous version. It's almost like the `require` cache is being used w/ hapi 17, but wasn't before.

vladikoff:
We need this continue working on the Hapi updates. I ran some tests and things seemed in order 👍

shane-tomlinson:
Hey, I forgot I had this PR open! Thanks @vladikoff!

PR #343, opened 2018-09-26
feat(profile): invalidate cache when profileChangedAt when needed

vbudhram:
Connects to https://github.com/mozilla/fxa-auth-server/issues/2490

This PR will invalidate a user's profile cache whenever auth-server reports a new `profileChangedAt` value (via oauth token), than what is stored in the cache it self.

### DIFF
--- a/lib/routes/_core_profile.js
+++ b/lib/routes/_core_profile.js
@@ -27,7 +27,8 @@ module.exports = {
       email: Joi.string().optional(),
       locale: Joi.string().optional(),
       amrValues: Joi.array().items(Joi.string().required()).optional(),
-      twoFactorAuthentication: Joi.boolean().optional()
+      twoFactorAuthentication: Joi.boolean().optional(),
+      profileChangedAt: Joi.number().optional()
     }
   },
   handler: function _core_profile(req, reply) {
@@ -83,6 +84,9 @@ module.exports = {
       }
       if (typeof body.authenticatorAssuranceLevel !== 'undefined') {
         result.twoFactorAuthentication = body.authenticatorAssuranceLevel >= 2;
+      }
+      if (typeof body.profileChangedAt !== 'undefined') {
+        result.profileChangedAt = body.profileChangedAt;
       }
       reply(result);
     });
--- a/lib/routes/profile.js
+++ b/lib/routes/profile.js
@@ -4,6 +4,7 @@
 
 const Joi = require('joi');
 const checksum = require('checksum');
+const P = require('../promise');
 
 const logger = require('../logging')('routes.profile');
 
@@ -37,10 +38,16 @@ module.exports = {
     const server = req.server;
     const creds = req.auth.credentials;
 
-    server.methods.profileCache.get(req, (err, result, cached, report) => {
+    function createResponse (err, result, cached, report) {
       if (err) {
         return reply(err);
       }
+
+      // `profileChangedAt` is an internal implementation detail that we don't
+      // return to reliers. As of now, we don't expect them to have any
+      // use for this.
+      delete result.profileChangedAt;
+
       if (creds.scope.indexOf('openid') !== -1) {
         result.sub = creds.user;
       }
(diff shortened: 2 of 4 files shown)


test/profileCache.js
```diff
@@ -207,6 +207,56 @@ describe('profile cache', function() {
...
+        })
+        .then(res => {
+          assert.ok(res.headers['last-modified']);
+          assert.equal(res.headers['last-modified'], lastModified);
+          mock.token({
+            user: userid,
+            scope: ['profile:avatar:write']
+          });
+          // upload avatar (should invaldate cache)
+          mock.image(imageData.length);
```
shane-tomlinson: I'm new to this repo too and am probably missing something obvious, this test looks the same as the previous test. Is the cache being invalidated because of the new code or because the avatar is uploaded?

vbudhram: Heh, no it is the same test, didn't get a chance to update it yet. I was going to use previous test as a basis for this one.

lib/routes/_core_profile.js
```diff
@@ -27,7 +27,8 @@ module.exports = {
       email: Joi.string().optional(),
       locale: Joi.string().optional(),
       amrValues: Joi.array().items(Joi.string().required()).optional(),
-      twoFactorAuthentication: Joi.boolean().optional()
+      twoFactorAuthentication: Joi.boolean().optional(),
+      profileChangedAt: Joi.number()
```
rfk: nit: we should explicitly annotate this as `.optional()` for consistency with the other fields.

lib/routes/_core_profile.js
```diff
@@ -84,6 +85,9 @@ module.exports = {
       if (typeof body.authenticatorAssuranceLevel !== 'undefined') {
         result.twoFactorAuthentication = body.authenticatorAssuranceLevel >= 2;
       }
+      if (typeof body.profileChangedAt !== 'undefined') {
+        result.profileChangedAt = body.profileChangedAt;
+      }
```
rfk: I'll add here as an early comment, we should probably treat this new field as an internal implementation detail and avoid returning it to reliers as part of the user's profile data.  We don't expect them to have any use for it.

vbudhram:
@mozilla/fxa-devs r?

test/profileCache.js
```diff
@@ -207,6 +210,41 @@ describe('profile cache', function() {
...
+        return makeProfileReq(userid);
+      })
+      .then(res => {
+        assert.ok(res.headers['last-modified']);
+        assert.equal(res.headers['last-modified'], lastModified);
+        return P.delay(500);
+      })
+      .then(() => {
+        // verify cache was invalidated due to profileChangedAt update
+        mock.profileChangedAt(MOZILLA_EMAIL);
```
rfk: Should this have `PROFILE_CHANGED_AT` as its second argument?  It looks like it might cause `profileChangedAt` to be undefined.

test/profileCache.js
```diff
@@ -207,6 +210,41 @@ describe('profile cache', function() {
...
+        return makeProfileReq(userid, undefined, PROFILE_CHANGED_AT_LATER_TIME);
+      })
+      .then(res => {
+        assert.ok(res.headers['last-modified']);
+        assert.notEqual(res.headers['last-modified'], lastModified);
```
rfk: It would be more accurate to test that `lastModified < res.headers['last-modified']` in the same way as the test below.

lib/routes/profile.js
```diff
@@ -60,7 +67,21 @@ module.exports = {
       } else {
         logger.info('batch.db');
       }
+
       return rep.header('last-modified', lastModified.toUTCString());
+    }
+
+    server.methods.profileCache.get(req, (err, result, cached, report) => {
+      // Check to see if the auth-server is reporting a newer `profileChangedAt`
+      // than the cached value. If so, lets invalidate the cache and set new value.
```
rfk: This comment could be a little misleading - it's not the auth-server that's reporting the newer timestamp in this case, it's the oauth-server from validating the token.

lib/routes/profile.js
```diff
@@ -60,7 +67,21 @@ module.exports = {
...
+      if (result && result.profileChangedAt < creds.profileChangedAt) {
```
rfk: It seems a bit weird to test for the existence of `result` here, since IIUC it will only be missing when there has been an error.  Would either an early return, or an `if (!err)` be clearer here?

